PR #1, opened 2019-06-11
constant paths as consts

HyamsG:


HyamsG:
Declare const as const.

PR #25, opened 2019-07-17
fix deprecation warning

erezalg:
fixed deprecation warnings in examples as well as tensorflow helper functions


erezalg:
@bmartinn , Check again

PR #134, opened 2020-05-16
address DeprecationWarning

erip:
Fixes DeprecationWarning: The usage of `cmp` is deprecated and will be removed on or after 2021-06-01.  Please use `eq` and `order` instead.


bmartinn:
Thanks @erip !
BTW, does `attr==18.0` has support for `eq` & `order` ?
Other than checking the minimum `attrs` version in the  requirements, nothing is blocking this PR :)

PR #143, opened 2020-06-05
remove six and pathlib2 dependencies from setup.py

### LINKED ISSUE #142: Error when installing from source: six and pathlib2 required
Because `setup.py` imports `six` and `pathlib2`, the installation of trains will fail in a new environment, because these packages are not installed. They should be installed by the user before installing trains/trains-agent, which is not intuitive not straightfoward IMO.

Both these dependencies can be removed with a very little refactoring effort, WDYT?

H4dr1en:
Fix #142

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,27 +3,31 @@ TRAINS - Artificial Intelligence Version Control
 https://github.com/allegroai/trains
 """
 
+import os.path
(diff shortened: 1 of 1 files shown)


bmartinn:
@H4dr1en  Awesome , thanks!

PR #145, opened 2020-06-15
Add task.wait_for_status()

H4dr1en:
This PR extends Task with the following function:

```
def wait_for_status(self, status=(tasks.TaskStatusEnum.completed),
                        raise_on_status=(tasks.TaskStatusEnum.failed),
                        check_interval_sec=60)
```

Use case: in a controller, it is handy to do as follows:

```
cloned_task = Task.clone(...)
cloned_task.connect(...)
cloned_task.enqueue(...)
cloned_task.wait_for_status(TaskStatusEnum.completed)
... # continue after task is finished
```

### DIFF
--- a/trains/task.py
+++ b/trains/task.py
@@ -13,7 +13,7 @@ try:
 except ImportError:
     from collections import Callable, Sequence as CollectionsSequence
 
-from typing import Optional, Union, Mapping, Sequence, Any, Dict, TYPE_CHECKING
+from typing import Optional, Union, Mapping, Sequence, Any, Dict, TYPE_CHECKING, Iterable
 
 import psutil
 import six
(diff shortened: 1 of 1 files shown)


bmartinn:
Hi @H4dr1en ,
Would it make more sense to have it as part of the [automation/TrainsJob](https://github.com/allegroai/trains/blob/master/trains/automation/job.py#L144) ?
At least when we think about controllers, the idea was to create a simple (optimized) interface to clone/enqueue/monitor a job , this was the reason to add the `TrainsJob` class.
It is also used by the [HyperParameterOptimizer](https://github.com/allegroai/trains/blob/master/trains/automation/optimization.py#L757) that actually implements a kind of complex controller.

It might makes sense to have it on the Task object as well, especially for completeness, 
but I suspect that for controllers we should have a simple interface like the TrainsJob.
What do you think?

H4dr1en:
> Would it make more sense to have it as part of the automation/TrainsJob ?

Hoo, very cool! I wasn't aware of this class, awesome! It would be nice to mention it somewhere in allegro.ai/docs. 

I agree, it makes more sense that it lives in TrainsJob!

bmartinn:
Yes absolutely, I'll make sure we have it in the documentation (probably will be with a section on controllers that we are working on :)

I think that what we will do is to merge the commit, and then have a similar functionality in TrainsJob (I especially like the `raise_on_status` :+1: )

H4dr1en:
Very cool, looking forward! Feel free to push any changes btw

PR #148, opened 2020-06-19
bug: test_loader was using train batch_size instead of test_batch_size

NumesSanguis:
`batch_size=args.batch_size` --> `batch_size=args.test_batch_size` for `test_loader`


bmartinn:
@NumesSanguis, Thank you for catching it  :)

PR #156, opened 2020-06-27
fix deprecated warnings

williamFalcon:
this likely applies to the rest of imports

### DIFF
--- a/trains/backend_interface/metrics/reporter.py
+++ b/trains/backend_interface/metrics/reporter.py
@@ -1,9 +1,9 @@
 import json
 
 try:
-    from collections.abc import Iterable
-except ImportError:
     from collections import Iterable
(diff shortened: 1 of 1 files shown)


jkhenning:
Hi @williamFalcon,

Thanks for contributing :)

A question - I'm not sure I understand the fix - I've tested the current code and your PR under conda 3.7, and I seem to get the `DeprecatinoWarning` when using the PR, but not when using the current code.

This seems to sit with the fact that in the current code, we first try `from collections.abc import Iterable`, which is the correct import for any version >2.7, and only use the deprecated form (`from collections import Iterable`) when that fails.

Can you please  explain which Python version / setup resulted in the deprecation warning using the current code?

allegroai-git:
Stale, closing.

PR #173, opened 2020-07-22
Add Optuna Integrated badge to README.md

Crissman:
Hi!

Thanks for reaching out to us about the integration of Optuna into Trains. Glad to see Optuna is helpful!

Here's a badge to show that Optuna is built in. If there are other ways to let allegro.ai users know about Optuna, let us know.

Thanks!


bmartinn:
Thanks @Crissman !

Now that Optuna is integrated with Trains hyper-parameter optimizer service, would it make sense to also add an example in the Optuna repo?

Here what I was thinking, the great combination of *Trains* and *Optuna* means that users can run the Optuna optimizer on **any** codebase without changing the code.

This all relies on the fact that once code is executed with *Trains*, all the environment setup is logged (git code base, python environment, argparser, etc.), then using the *trains-agent* you can launch multiple training sessions with different parameters, and monitor their results from an external machine (e.g. the controller).

This means you can do multi-node optimization with a click of a button (with the help of [trains-agent](https://github.com/allegroai/trains-agent) and Optuna acting as controller). 

What do you think?

PR #184, opened 2020-08-11
Only install typing dependency on Python<3.5

AdrianoKF:
Installing `typing` on recent Python versions leads to the error described in python/typing#573 (running tests with tox yields `AttributeError: type object 'Callable' has no attribute '_abc_registry'`).

The [`typing` readme](https://pypi.org/project/typing/) suggests to install the package only for Python versions <3.5, which this commit does.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,5 +20,5 @@ requests-file>=1.4.2
 requests>=2.20.0
 six>=1.11.0
 tqdm>=4.19.5
-typing>=3.6.4
+typing>=3.6.4 ; python_version < '3.5'
 urllib3>=1.21.1


bmartinn:
Thanks @AdrianoKF !
I can't remember I witnessed any issue with `typing` in python >= 3.6 ...
Is this a bug in a specific `typing` version ? 
(Not objecting to the PR, just trying to better understand the root cause)

AdrianoKF:
Thanks for your quick response, @bmartinn!

Unfortunately, I don't know the exact root cause - and afaict none was given from the upstream issue python/typing#573. It seems that the Python-bundled version of `typing` (available since Python 3.6) seems to clash with the pip-installed version in some edge cases, leading to the `AttributeError` I mentioned. I was trying to reproduce the issue with a colleague, who didn't see it on his machine, but I was consistently unable to run tox in our project which includes trains on a CI build on the Docker `python:3.8-buster` base image, which is about as vanilla as it gets :)

If you'd like, we can try and reduce our issue down to a minimal reproducible example, but given that the typing library itself explicitly mentions that installing the on Python > 3.5 has no effect (or even a detrimental one, as evidenced by the issue), the change is probable a sensible one.

jkhenning:
Hi @AdrianoKF,

From what I can see this is indeed the best practice, and shouldn't change the current behavior from Trains' point of view :slightly_smiling_face: 

Thanks for this contribution :+1:

PR #226, opened 2020-10-23
Enable re-downloading cached files

eliorc:
# Summary

This PR introduces a flag for the `StorageManager.get_local_copy` named `force_download` which when set to `True` will overwrite existing cached files (default to `False`).
This flag is relevant for cases where the user modifies the local copy without uploading it, or when a file is updated on the remote repository and the user wishes to re-download it.

## New features

 - `force_download` flag on `StorageManager.get_local_copy`.

## Other changes & Fixes
 
 - Bad usage of `Optional` type, `Optional` is for arguments that explicitly allow the value `None` and not for optional arguments as intuition might suggest ([reference](https://docs.python.org/3/library/typing.html#typing.Optional))
 - Deletion of old files was dependent on modification time rather than access time, even though the sorting function was named `sort_max_access_time` in fact `stat().st_mtime` was used.
 - Switched the `.touch()` that updates access time + modification time to a new method called `update_access_time` to only update access time. This is relevant for potential future work that will be dependent on modification times

### DIFF
--- a/trains/storage/cache.py
+++ b/trains/storage/cache.py
@@ -27,7 +27,7 @@ class CacheManager(object):
             self._file_limit = max(self._file_limit, int(cache_file_limit))
             return self._file_limit
 
-        def get_local_copy(self, remote_url):
+        def get_local_copy(self, remote_url, force_download):
             helper = StorageHelper.get(remote_url)
             if not helper:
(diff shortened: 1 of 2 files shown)


eliorc:
Removed the commit containing the code changing the use of modification time to access time as a criteria for deleting old files

bmartinn:
LGTM :)

Issue #194, opened 2020-08-25
task.upload_artifact to wait until upload is finished before returning


H4dr1en:
Slack thread > https://allegroai-trains.slack.com/archives/CTK20V944/p1598346720094900

This is a proposal to extend `Task.upload_artifact` as follows:
```
def upload_artifact(..., wait_for_upload: bool = False):
    ...
    if wait_for_upload:
        self.flush(wait_for_uploads=True)
```

bmartinn:
Thanks @H4dr1en !
I'll update here once we add the new functionality.
(My hope is to have better granularity, so you are not waiting for all uploads of the system, but specifically to the current artifact)

allegroai-git:
Closing since related PR was merged and released ClearML/Trains v0.16.3.
Thanks for the PR! 😄

PR #231, opened 2020-10-27
Added synchronous support for upload_artifact()

OmerM25:
Solving #194

### DIFF
--- a/trains/binding/artifacts.py
+++ b/trains/binding/artifacts.py
@@ -305,7 +305,7 @@ class Artifacts(object):
         self.flush()
 
     def upload_artifact(self, name, artifact_object=None, metadata=None, preview=None,
-                        delete_after_upload=False, auto_pickle=True):
+                        delete_after_upload=False, auto_pickle=True, wait_on_upload=False):
         # type: (str, Optional[object], Optional[dict], Optional[str], bool, bool) -> bool
         if not Session.check_min_api_version('2.3'):
             LoggerRoot.get_base_logger().warning('Artifacts not supported by your TRAINS-server version, '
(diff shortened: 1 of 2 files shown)


trains/binding/artifacts.py
```diff
@@ -701,11 +705,14 @@ def _upload_local_file(
                          override_filename=override_filename,
                          override_filename_ext=override_filename_ext,
                          override_storage_key_prefix=self._get_storage_uri_prefix())
-        _, uri = ev.get_target_full_upload_uri(upload_uri)
+        _, uri = ev.get_target_full_upload_uri(upload_uri, quote_uri=False)
```
jkhenning, line 708: Are you not quoting the `uri` since that's what's required by `StorageManager`? In any case, these method returns the `uri`, so in order to maintain compatibility it has to be quoted before being returned

OmerM25: The `quote_uri=False` is because when you upload a file with spaces in the name, it is already being translated (so that " " -> "%20") before `ev.get_target_full_upload_uri()`, which will then translate the "%20" to "%2520" (translating the '%' as a character) which causes the file to be uploaded to a different uri than the one it shows on the ui.

OmerM25: @jkhenning Any thoughts?

jkhenning: @OmerM25 sorry for the delay... 
I get what you're saying, it's just that by-design, this method returns the proper URL, which now will not be quoted.
I think the correct thing to do is to add another call to `get_target_full_upload_uri()` in order to generate a valid URI to return, like so:
```python
...
if wait_on_upload:
  StorageManager.upload_file(local_file, uri)
else:
  self._task._reporter._report(ev)

_, quoted_uri = ev.get_target_full_upload_uri(upload_uri, quote_uri=True)
return quoted_uri
```

OmerM25: That makes more sense @jkhenning 
I've uploaded the fix, thank you!

PR #237, opened 2020-11-09
Add cifar ignite example and add auto extract of tar.gz files when using storagemanager

erezalg:
Also fixed a bug in manager.py in _extract_to_cache function where Null name variable would cause crash when calling 
name = encode_string_to_filename(name)

### DIFF
--- a/trains/storage/manager.py
+++ b/trains/storage/manager.py
@@ -1,5 +1,5 @@
-import os
 import shutil
+import tarfile
 from random import random
 from time import time
 from typing import Optional
@@ -7,10 +7,10 @@ from zipfile import ZipFile
 
 from pathlib2 import Path
 
+from .cache import CacheManager
 from .util import encode_string_to_filename
 from ..debugging.log import LoggerRoot
-from .cache import CacheManager
-import tarfile
+
 
 class StorageManager(object):
     """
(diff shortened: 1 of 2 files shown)


trains/storage/manager.py
```diff
@@ -97,13 +97,16 @@ def _extract_to_cache(cls, cached_file, name):
         :return: cached folder containing the extracted archive content
         """
         # only zip files
-        if not cached_file or not str(cached_file).lower().endswith('.zip'):
+        if not cached_file or (not str(cached_file).lower().endswith('.zip')\
```
jkhenning, line 100: Please change to:
```python
if not (cached_file and any(map(str(cached_file).lower().endswith, (".zip", ".tar.gz"))):
  ...
```

trains/storage/manager.py
```diff
@@ -97,13 +97,16 @@ def _extract_to_cache(cls, cached_file, name):
...
+                and not str(cached_file).lower().endswith('.tar.gz')):
             return cached_file
 
         cached_folder = Path(cached_file).parent
-        archive_suffix = cached_file.rpartition(".")[0]
-        name = encode_string_to_filename(name)
-        target_folder = Path("{0}_artifacts_archive_{1}".format(archive_suffix, name))
+        archive_prefix = cached_file.rpartition(".")[0]
```
jkhenning, line 105: Please change to:
```python
cached_file = Path(cached_file)
cached_folder = cached_file.parent
name = encode_string_to_filename(name) if name else name
target_folder = Path("{0}_artifacts_archive_{1}".format(cached_file.stem, name))
...
if cached_file.suffix == '.zip':
  ZipFile(str(cached_file)).extractall(path=temp_target_folder.as_posix())
elif cached_file.suffixes == ['.tar', '.gz']:
  with tarfile.open(str(cached_file)) as file:
    file.extractall(temp_target_folder)
```

Make sure to change further references to `cached_file` to `str(cached_file)`

Issue #250, opened 2020-11-18
Ligthgbm bind error with trains==0.16.3.


noklam:
Error

Reproducible script:
https://gist.github.com/noklam/0a1d261318621227c3bac30a50a0301b

```
  File "C:\Users\CHANNO.OOCLDM\AppData\Local\Continuum\anaconda3\envs\cba\lib\site-packages\trains\binding\frameworks\lightgbm_bind.py", line 122, in _train
    params, train_set = args
ValueError: not enough values to unpack (expected 2, got 0)
make: *** [Makefile:20: debug] Error 1
```

allegroai-git:
Closing this as the PR was merged and released in ClearML/Trains v0.16.4
Thanks for the PR! 😄

PR #251, opened 2020-11-18
Fix lightgbm binding keyword argument bugs

noklam:
Prevent error when user using keyword arguments instead of positional arguments
Related: #250 

lgb.train support positional arguments, but user can pass it as a keyword arguments. Previously code will fail if args is empty.

This fix attempt to catch positional arguments into `kwargs` if exist, so it can support both positional/keyword arguments.

### DIFF
--- a/trains/binding/frameworks/lightgbm_bind.py
+++ b/trains/binding/frameworks/lightgbm_bind.py
@@ -118,11 +118,19 @@ class PatchLIGHTgbmModelIO(PatchBaseModelIO):
                 except Exception:
                     pass
             return callback
-        params, train_set = args
+        
+        if args:
+            kwargs['params'] = args.pop(0)
+        if args:
+            kwargs['train_set'] = args.pop(0)
+        if args:
+            raise ValueError('It only supports 2 position argument {params} and {trian_set}')
+
         kwargs.setdefault("callbacks", []).append(trains_lightgbm_callback())
-        ret = original_fn(params, train_set, **kwargs)
+        ret = original_fn(**kwargs)
         if not PatchLIGHTgbmModelIO.__main_task:
             return ret
+        params = kwargs['params']
         for k, v in params.items():
             if isinstance(v, set):
                 params[k] = list(v)
--- a/trains/binding/frameworks/lightgbm_bind.py
+++ b/trains/binding/frameworks/lightgbm_bind.py
@@ -120,11 +120,11 @@ class PatchLIGHTgbmModelIO(PatchBaseModelIO):
             return callback
         
         if args:
-            kwargs['params'] = args.pop(0)
-        if args:
-            kwargs['train_set'] = args.pop(0)
-        if args:
-            raise ValueError('It only supports 2 position argument {params} and {trian_set}')
+            kwargs['params'] = args[0]
+            if len(args) == 2:
+                kwargs['train_set'] = args[1]
+            elif len(args) > 2:
+                raise ValueError('It only supports 2 position argument {params} and {trian_set}')
 
         kwargs.setdefault("callbacks", []).append(trains_lightgbm_callback())
         ret = original_fn(**kwargs)
--- a/trains/binding/frameworks/lightgbm_bind.py
+++ b/trains/binding/frameworks/lightgbm_bind.py
@@ -119,18 +119,11 @@ class PatchLIGHTgbmModelIO(PatchBaseModelIO):
                     pass
             return callback
         
-        if args:
-            kwargs['params'] = args[0]
-            if len(args) == 2:
-                kwargs['train_set'] = args[1]
-            elif len(args) > 2:
-                raise ValueError('It only supports 2 position argument {params} and {trian_set}')
-
         kwargs.setdefault("callbacks", []).append(trains_lightgbm_callback())
-        ret = original_fn(**kwargs)
+        ret = original_fn(*args, **kwargs)
         if not PatchLIGHTgbmModelIO.__main_task:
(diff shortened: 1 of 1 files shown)


trains/binding/frameworks/lightgbm_bind.py
```diff
@@ -118,11 +118,19 @@ def callback(env):
...
+        if args:
+            kwargs['params'] = args.pop(0)
+        if args:
+            kwargs['train_set'] = args.pop(0)
+        if args:
+            raise ValueError('It only supports 2 position argument {params} and {trian_set}')
+
         kwargs.setdefault("callbacks", []).append(trains_lightgbm_callback())
-        ret = original_fn(params, train_set, **kwargs)
+        ret = original_fn(**kwargs)
```
JDennisJ, line 130: How about sending the args to the original function too? Just `ret = original_fn(*args, **kwargs)`, and let's not parse the args at all.

With this parsing method, we will get an exception for using `pop` (and basically, we need to get only the `params`).

trains/binding/frameworks/lightgbm_bind.py
```diff
@@ -118,11 +118,19 @@ def callback(env):
...
         if not PatchLIGHTgbmModelIO.__main_task:
             return ret
+        params = kwargs['params']
```
JDennisJ, line 133: For getting the `params`, we can use `params = args[0] if args else kwargs['params']` to support both ways

JDennisJ (changes requested):

trains/binding/frameworks/lightgbm_bind.py
noklam: Why would it get exception? if args is empty, it would not trigger the pop() at all. train_set is also a positional argument, so we need to have to deal with 3 condition
1. both positional
2. params positional and train_set is keyword
3. both keywords

noklam: if args is an empty list  -> *args will become *None and will throw error instead?

trains/binding/frameworks/lightgbm_bind.py
noklam: This can be used if pop() is not used, but I found using pop keep it cleaner and I can get rid of checking is arg is None or leng == 1 or len ==2

trains/binding/frameworks/lightgbm_bind.py
JDennisJ: `args` are `Tuple`, so `pop` raising an exception (for conditions 1 and 2).

What do you get running https://github.com/allegroai/trains/blob/master/examples/frameworks/lightgbm/train_with_lightbgm.py (condition 1)?

trains/binding/frameworks/lightgbm_bind.py
JDennisJ: https://github.com/allegroai/trains/pull/251/files#r526154397, we can not use `pop`...

trains/binding/frameworks/lightgbm_bind.py
noklam: @JDennisJ How about the fix now. should work for all 3 cases.

JDennisJ: For those 3 cases - yes, but with this solution, you are guessing the user will send a max of 2 parameters for `args` (or forcing the user to do so).

If we will take the example above (https://github.com/allegroai/trains/blob/master/examples/frameworks/lightgbm/train_with_lightbgm.py), and replace [this](https://github.com/allegroai/trains/blob/master/examples/frameworks/lightgbm/train_with_lightbgm.py#L50) with

```
gbm = lgb.train(
    params,
    lgb_train,
    20,
    valid_sets=lgb_eval,
    early_stopping_rounds=5
)
```
it should still work, but we will get an exception with the changes.

But, if we will send `ret = original_fn(*args, **kwargs)`, we are just a pipe in the way, so if the user used `lgb.train` in any valid way, we will pass it to the original function and won't fail.

Other than that, we are parsing `train_set` without any reason.

noklam (commented):
@JDennisJ I modified the PR. Yes you are right, now I don't even know why I thought I had to parse the argument in the first place, it seems obvious.

trains/binding/frameworks/lightgbm_bind.py
```diff
@@ -118,11 +118,12 @@ def callback(env):
...
                     pass
             return callback
-        params, train_set = args
+        
         kwargs.setdefault("callbacks", []).append(trains_lightgbm_callback())
-        ret = original_fn(params, train_set, **kwargs)
+        ret = original_fn(*args, **kwargs)
         if not PatchLIGHTgbmModelIO.__main_task:
             return ret
+        params = args[0] if args else kwargs['params']
```
JDennisJ, line 126: `kwargs['params']` -> `kwargs.get('params', {})` just to be on the safe side

JDennisJ: and need to add [here](https://github.com/allegroai/trains/pull/251/files#diff-7419bbefd8f631446dbf82d2e6bae476070d71d5735a27791aeca2b80218bfb4L129) `if params` just for the `connect` part

noklam: The PR is merged already, can you make these changes directly, or I should create another PR?

JDennisJ: @noklam Sure, I will make those.

Thanks for your contribution  :)

PR #266, opened 2020-12-18
enabling custom AWS host value

valeriano-manassero:
This proposal enables the option to set custom host for AWS Storage manager by env var.

In this way it will be possible to add this configurable parameter in Helm chart enabling declaration of any external non-AWS S3 (e.g. Minio) directly from Helm release.

### DIFF
--- a/trains/backend_config/bucket_config.py
+++ b/trains/backend_config/bucket_config.py
@@ -89,13 +89,14 @@ class BaseBucketConfigurations(object):
 
 class S3BucketConfigurations(BaseBucketConfigurations):
     def __init__(
-        self, buckets=None, default_key="", default_secret="", default_region=""
+        self, buckets=None, default_key="", default_secret="", default_region="", default_host=""
     ):
         super(S3BucketConfigurations, self).__init__()
         self._buckets = buckets if buckets else list()
         self._default_key = default_key
         self._default_secret = default_secret
         self._default_region = default_region
+        self._default_host = default_host
         self._default_multipart = True
 
     @classmethod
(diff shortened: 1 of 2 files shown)


jkhenning:
Hi @valeriano-manassero,

I'm not sure I understand the use-case. The `host` parameter in each bucket configuration, if provided, is used by Trains when trying to locate the correct bucket configuration for the provided URL.

The default values are used when no specific configuration is provided. However, even when no specific configuration is provided, you're still trying to download/upload a specific URL, which usually contains a host in case the bucket in question is not located in AWS S3 (if it is, and a host is not provided, boto3 will determine the host using the provided `region`).

So, as far as I can see, the only use-case is allowing Trains to deduce a non-AWS host when a URI is provided with no host (e.g. `s3://my-bucket`) - a URI which as a standard is used to indicate an AWS host.

If so, this is more than an override value than a default value, and should be named as such :)
Also, I think in this case you're missing the actual usage, i.e. use this value when obtaining credentials for a URI (see `bucket_config.py` in the `get_config_by_uri()` method, around line `191` - the current behavior is to try and locate a port number in the URI, and assume that if a port number exists it means we're dealing with a non-AWS host, otherwise it is an AWS host. An `override_host` value will mean simply setting the host there and not trying to find the port number at all...

What do you think?

valeriano-manassero:
While I was testing many different configurations, I came up with this implementation based on wrong assumptions like you said. In fact I have another problem with Minio but it is not related to host like I was thinking at the start of the process (I will open an issue soon).

So I agree with you this implementation is not bringing a clear improvement and I'm going to drop it.
Thank you for your feedback!

PR #267, opened 2020-12-23
fix task init command in readme

harrywang:
Change `task = Task(project_name='examples', task_name='hello world')` to `task = Task.init(project_name='examples', task_name='hello world')` to get rid of the following error:
```
clearml.errors.UsageError: Task object cannot be instantiated externally, use Task.current_task() or Task.get_task(...)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ pip install clearml
 Add two lines to your code:
 ```python
 from clearml import Task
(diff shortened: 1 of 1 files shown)


bmartinn:
Thanks @harrywang !
Greatly appreciated :)

Issue #294, opened 2021-01-22
remote diff should not be stripped


tankeco:
remote diff is stripped here:
https://github.com/allegroai/clearml/blob/master/clearml/backend_interface/task/repo/detectors.py#L142

This causes the `git apply` sometimes fail. 

In contrast, the processing is correct when it is not remote diff:
https://github.com/allegroai/clearml/blob/master/clearml/backend_interface/task/repo/detectors.py#L126

bmartinn:
Hi @tankeco 
No need to be shy, if you find a bug feel free to just PR a fix :)

Regrading the specific issue, in both cases the results from git commands are stripped only if they are Not `diff` (in the remote case it includes `diff_remote` and `diff_remote_fallback`), see Commands [here](https://github.com/allegroai/clearml/blob/715a0d7edc33b6ff91affe525194fa0c525d98f3/clearml/backend_interface/task/repo/detectors.py#L57)
Is there a specific scenario it fails?

bmartinn:
Closing after the fix was meged

PR #295, opened 2021-01-23
fix strip remote diff

tankeco:
the results from git commands should be stripped only if they are **Not** diff (in the remote case it includes diff_remote and diff_remote_fallback)
see https://github.com/allegroai/clearml/issues/294


bmartinn:
Hi @tankeco 

Git Diff's should not be stripped, it used to be the case, but it broke compatibility between Windows and Linux git diff. 

If I remember correctly, if you have \n at the end of a file (for some reason it can happen on Windows more often than on Linux) then stripping the git diff will essentially break the git apply later.

WDYT?

bmartinn (approved):

PR #329, opened 2021-04-06
Update README.md to include community roadmap

stelford:
Initial wild stab at community roadmap, obviously open to lots of changes :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,15 +147,14 @@ Additionally, you can always find us at *clearml@allegro.ai*
 
 ## Community Roadmap
 
-The upcoming community release 0.18 has the following keypoints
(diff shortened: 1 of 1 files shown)


bmartinn:
Hi @stefempath 

I think that with "release 0.18" we actually refer to the next `cleaml` python package release, assuming this is 0.17.6, correct ?

I think that "Fix reporting after task closing" is just a bug fix, not really road-map worthy, no?

stelford:
Take #2 on the verbage and item list :)

bmartinn:
I would push the documentation to v1.0
Also, we are adding multi-model per Task support in v1.0

stelford:
I will need to close this one down and re-open against the Tag 1.0 :)

PR #357, opened 2021-05-12
Fix changes in Hydra binding for upcoming version.

ramonemiliani93:
For compatibility in the upcoming Hydra change we update the binding to use keyword arguments when calling `run_job`.


ramonemiliani93:
@bmartinn Let me know if it needs more changes! I tested it locally and it works with the new Hydra version 👌

PR #372, opened 2021-05-31
Add a tensorboardx example for add_video (creates GIFs in tensorboard)

mctigger:
Currently, `writer.add_video("my_video", video, 0, fps=50)` produces broken still images. This is an example to facilitate support of `add_video`. When run you should see an video/gif that simply fades black to white and back in a loop.


bmartinn:
Looks good.
I'll update here once we have fix/support pushed.
(Internal tests already pass, I'm hoping to see the code public in a couple of days)

mctigger:
Tried and works fine! Perfect!

PR #384, opened 2021-06-17
Get notebook auth token from the JUPYTERHUB_API_TOKEN environment variable

mallman:
Get notebook auth token from the JUPYTERHUB_API_TOKEN environment variable if it is not present in the server info.

From our experience, notebooks spawned by JupyterHub do not provide an API token in their server info. In these cases, the clearml client is unable to integrate with Jupyter. This PR adds a fallback option to check the JUPYTERHUB_API_TOKEN environment variable if it can't find the API token otherwise. With this capability, we are able to integrate with notebooks spawned by a JupyterHub spawner.


bmartinn:
Thanks @mallman 
Looks straight forward to me :)
👍

PR #456, opened 2021-09-20
Fix xgb train overloading

jalexand3r:
Per discussion in dev slack on 9/20/21, this fixes an issue that causes an error when xgboost is trained using keyword arguments.

For example, if [the example code](https://github.com/allegroai/clearml/blob/master/examples/frameworks/xgboost/xgboost_sample.py#39) is changed from
`bst = xgb.train(param, dtrain, num_round)`
to 
`bst = xgb.train(params=param, dtrain=dtrain, num_boost_round=num_round) `
We receive the following error:  `TypeError: _train() missing 1 required positional argument: '_PatchXGBoostModelIO__obj'`

This minor change fixes that issue.

### DIFF
--- a/clearml/binding/frameworks/xgboost_bind.py
+++ b/clearml/binding/frameworks/xgboost_bind.py
@@ -116,13 +116,13 @@ class PatchXGBoostModelIO(PatchBaseModelIO):
         return model
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


bmartinn:
Thanks @jalexand3r !
Looking good, let's merge.

PR #462, opened 2021-09-30
Add LogUnifomParameterRange for hyperparameter optimization with optuna

diegopereyra99:
As discussed in slack clearml-community channel, implemented a logarithmic parameter range for hyperparamter optimization.

### DIFF
--- a/clearml/automation/parameters.py
+++ b/clearml/automation/parameters.py
@@ -171,6 +171,19 @@ class UniformParameterRange(Parameter):
         return [{self.name: v} for v in values]
 
 
+class LogUniformParameterRange(UniformParameterRange):
+    def __init__(self, name, min_value, max_value, step_size=None, include_max_value=True):
+        super().__init__(name, min_value, max_value, step_size=step_size, include_max_value=include_max_value)
+    
+    def get_value(self):
+        values_dict = super().get_value()
+        return {self.name: self.base**v for v in values_dict.values()}
+    
+    def to_list(self):
+        values_list = super().to_list()
+        return [{self.name: self.base**v[self.name]} for v in values_list]
+    
+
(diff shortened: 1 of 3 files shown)


clearml/automation/parameters.py
```diff
@@ -171,6 +171,19 @@ def to_list(self):
         return [{self.name: v} for v in values]
 
 
+class LogUniformParameterRange(UniformParameterRange):
+    def __init__(self, name, min_value, max_value, step_size=None, include_max_value=True):
+        super().__init__(name, min_value, max_value, step_size=step_size, include_max_value=include_max_value)
+    
+    def get_value(self):
+        values_dict = super().get_value()
+        return {self.name: self.base**v for v in values_dict.values()}
```
jkhenning, line 180: @diegopereyra99 I don't see `self.base` defined anywhere?

diegopereyra99: I had defined `self.base` in the `__init__` but then I removed it since the Optuna Optimizer (then one I'm using) expects de min and max values to be logarthmic. I overrode the `get_value` and `to_list` methods as @bmartinn suggested in slack but then I did not used them. The problem here is that `min_value` and `max_value` are supposed to be from a linear space that is then converted to a logarthmic one, but this would be different when using optuna.

bmartinn: @diegopereyra99 I think that we should set `self.base=e` , there is no need to worry about using it as in the `optuna` integration layer we simply ignore it and pass the original min/max values.
This way if the Class is used with Grid/Random search algorithms, we still get the random values in log-scale.
Sounds okay?

jkhenning: Hey @diegopereyra99 ,

Looks good! I'll merge :)

Issue #514, opened 2021-12-14
Add better input handling for clearml-init in colab


thepycoder:
The ipykernel cannot yet handle input requests coming from a terminal command ran by using the `!` operator, but Google colab can!
https://github.com/ipython/ipykernel/issues/513

Currently, the docs suggest manually setting the credentials using `Task.set_credentials`, because colab input handles input a little different and so the default prompt using `clearml-init` is not working properly.

## Simple example
Go to https://colab.research.google.com/
Start a new notebook
Install clearml using `!pip install clearml`
In the next cell, run `!clearml-init`
Copy paste credentials in the prompt, they will be squished all in 1 line.
It will keep prompting for new lines because it expects multi-line input, not all in 1 line.

## Expected Behaviour
I should be able to copy paste the credentials in the input field and have it be registered correctly

## Possible fix
I have a fix working over at https://github.com/thepycoder/clearml/commit/4d57beaa3ddb37bc444e587365c27603953b2d27 that can fix this by adding a `--colab` flag to the `clearml-init` command. 
However, it might be more user friendly to look in the environment variables for `COLAB_GPU` which should always be set (either to 0 or 1) and trigger the flag that way, but could be annoying when this env var is set in a local environment (which seems highly unlikely to me)

What do you think?

jkhenning:
Thanks @thepycoder! PR merged, closing this issue 🙂

PR #515, opened 2021-12-14
Add better input handling for clearml-init in colab

thepycoder:
It's specifically for use in Google Colab. A Colab instance always has the `COLAB_GPU` environment variable (this only one that's distinctive afaik). It even has the env var when you have a CPU or TPU runtime, in those cases it just says `COLAB_GPU=0`.

Fixes issue #514


jkhenning:
Cool! Thanks 🙂

PR #521, opened 2021-12-22
Add environment variable for request method

### LINKED ISSUE #517: Bad Request when authenticating to ClearML server
### Summary

I'm getting a bad request error (400) when trying to connect to a self hosted server via the python API. I've traced it down to the `Content-Length` header. If I set it to `0` or remove it the request is successful, but otherwise it is set to 2 and the request fails. I believe the root cause is this line:

https://github.com/allegroai/clearml/blob/fd2d6c6f5d46cad3e406e88eeb4d805455b5b3d8/clearml/backend_api/session/session.py#L710

Replacing the default argument with `None` like below ensures that the `Content-Length` header is set correctly and works for the request in question, but seems to break other things (I'll keep looking).

```python
data = {"expiration_sec": exp} if exp else None
```

Unfortunately the error is not repeatable on the ClearML community server (I'm not sure why!) so hard to provide a MWE.

### What I did

Set the environment variables `CLEARML_API_ACCESS_KEY`, `CLEARML_API_SECRET_KEY`, `CLEARML_API_HOST`, `CLEARML_WEB_HOST`, `CLEARML_FILES_HOST` to connect to my self-hosted server (which is running v1.1.0 as 1.1.1 is not available in the GCP Images yet).

Attempt to connect via:

```python
from clearml import Task
project = "ProjectName"
Task.get_tasks(project_name=project))
```

And get this error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/task.py", line 847, in get_tasks
    return cls.__get_tasks(task_ids=task_ids, project_name=project_name, tags=tags,
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/task.py", line 3535, in __get_tasks
    for task in cls._query_tasks(project_name=project_name, task_name=task_name, **kwargs)]
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/task.py", line 3553, in _query_tasks
    cls._get_default_session(),
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/backend_interface/base.py", line 101, in _get_default_session
    InterfaceBase._default_session = Session(
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/backend_api/session/session.py", line 201, in __init__
    self.refresh_token()
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/backend_api/session/token_manager.py", line 112, in refresh_token
    self._set_token(self._do_refresh_token(self.__token, exp=self.req_token_expiration_sec))
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/backend_api/session/session.py", line 739, in _do_refresh_token
    six.reraise(*sys.exc_info())
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/six.py", line 719, in reraise
    raise value
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/experiment-test-qPFpYaQg-py3.9/lib/python3.9/site-packages/clearml/backend_api/session/session.py", line 725, in _do_refresh_token
    raise LoginError(
clearml.backend_api.session.session.LoginError: Failed getting token (error 400 from https://api.cleaml.mydomain.com): Bad Request
```

### Setup

```
clearml==1.1.4
$ python --version
Python 3.9.9
```

mmiller-max:
Following discussion in #517 , this PR checks the environment variable `CLEARML_HTTP_REQUEST_METHOD` with a default of `get` when refreshing the login token and in `Request`. This allows `POST` requests to be used in situations where `GET` requests are not allowed to have payloads (e.g. some elastic load balancers).

Couple of questions:

1. I've just used `os.environ` directly, but instead we could use an `EnvEntry` like the others in `defs.py`. Which way would be preferred?
2. No validation is being done on whether the method in the environment variable is suitable (probably just `get` and `post`). Would this be preferable or we happy that if a user is adequately warned when they set this variable then that's sufficient?
3. It would be useful for this to be documented, but I can't see where this is done in the repository
4. I've tested it locally, but do we need some formal tests for this?
5. Is the name `CLEARML_HTTP_REQUEST_METHOD` too generic?

Closes #517


jkhenning:
@mmiller-max for the sake of verbosity, I would prefer `CLEARML_API_DEFAULT_REQ_METHOD`

jkhenning:
As for the other questions:

1. It would indeed be nice to have it defined there 🙂 
2. Yeah, I guess a single point of validation (can be done in the global context of `request.py`)
3. We have a `clearml-docs` repository for that, although I'm really not sure where this should appear... I think a good start if having it in the `defs.py` file with a good docstring (which includes some disclaimer explaining why this is "experimental")
4. Assuming it looks safe (i.e. no change in behavior if not set), I'll be happy to merge it 🙂

mmiller-max:
Let me know what you think of the docstring, would you rather it be in the module docstring for `defs` or is how I've done it OK?

clearml/backend_api/session/request.py
```diff
@@ -1,15 +1,22 @@
...
 
 from .apimodel import ApiModel
 from .datamodel import DataModel
+from .defs import ENV_API_DEFAULT_REQ_METHOD
 
 
-class Request(ApiModel):
-    _method = 'get'
+if ENV_API_DEFAULT_REQ_METHOD.get().upper() not in ("GET", "POST"):
+        raise ValueError(
```
jkhenning, line 13: Please reduce indentation to 4 spaces 🙂

clearml/backend_api/session/defs.py
```diff
@@ -17,3 +17,12 @@
...
 ENV_ENABLE_FILES_CONFIG_SECTION = EnvEntry('CLEARML_ENABLE_FILES_CONFIG_SECTION', type=bool)
+
+"""
+Experimental option to set the request method for all API requests and auth login.
+This could be useful when GET requests with payloads are blocked by a server as
+POST requests can be used instead.
+
+However this has not been vigorously tested and may have unintended consequences.
+"""
+ENV_API_DEFAULT_REQ_METHOD = EnvEntry("CLEARML_API_DEFAULT_REQ_METHOD", default="get")
```
jkhenning, line 28: I would much rather have no default value and have the default value stated explicitly when calling `get()`

clearml/backend_api/session/request.py
mmiller-max: Not sure what happened here .. !

clearml/backend_api/session/defs.py
```diff
@@ -17,3 +17,12 @@
...
 ENV_ENABLE_FILES_CONFIG_SECTION = EnvEntry('CLEARML_ENABLE_FILES_CONFIG_SECTION', type=bool)
+
+"""
+Experimental option to set the request method for all API requests and auth login.
+This could be useful when GET requests with payloads are blocked by a server as
+POST requests can be used instead.
+
+However this has not been vigorously tested and may have unintended consequences.
+"""
+ENV_API_DEFAULT_REQ_METHOD = EnvEntry("CLEARML_API_DEFAULT_REQ_METHOD")
```
jkhenning, line 28: Missing new line 🙂

clearml/backend_api/session/request.py
```diff
@@ -1,15 +1,22 @@
...
 import six
 
 from .apimodel import ApiModel
 from .datamodel import DataModel
+from .defs import ENV_API_DEFAULT_REQ_METHOD
 
 
-class Request(ApiModel):
-    _method = 'get'
+if ENV_API_DEFAULT_REQ_METHOD.get().upper() not in ("GET", "POST"):
```
jkhenning, line 12: Are you sure this won't fail? If `ENV_API_DEFAULT_REQ_METHOD` has no default value, I think this should be converted to:
```python
if ENV_API_DEFAULT_REQ_METHOD.exists() and ENV_API_DEFAULT_REQ_METHOD.get().upper() not in ("GET", "POST"):
  ...
```

mmiller-max: Yes sorry, just missed the default here. I think your method is more explicit here where the variable is being checked, than `ENV_API_DEFAULT_REQ_METHOD.get(default="get").upper() ` which is good.

mmiller-max:
@jkhenning do you think this change can make it into `1.1.5`?

jkhenning:
Yeah, it will 🙂

mmiller-max:
Perfect, thanks!

Issue #520, opened 2021-12-21
Run HPO completely locally


thepycoder:
When running an HPO task, I can run the optimizer task itself in my local environment, but I seem to have no choice but to run a separate agent to actually process the resulting training tasks.
This seems like overkill for someone without access to multiple machines and just adds overhead by recreating the environment again.

There already exists a `LocalClearmlJob` job type that we could use for this functionailty.
https://github.com/allegroai/clearml/blob/fd2d6c6f5d46cad3e406e88eeb4d805455b5b3d8/clearml/automation/job.py#L573

So maybe we should have it as an argument when calling `optimizer.start()`? That would make most sense.

I'm starting work on a PR for this.

jkhenning:
Done

PR #525, opened 2021-12-27
Add LocalClearmlJob as possible option for HPO

thepycoder:
Solves issue #520

Opted for a `run_locally()` method instead of adding the option as an argument to .run() because it is more inline with how ClearML pipelines manage the same functionality.

https://github.com/allegroai/clearml/blob/master/clearml/automation/controller.py#L703

-> Added the method, copied most of the docstring from .start() method, it also handles the same arguments
-> Changed LocalClearmlJob docstring to not only include the debugging usecase
-> Added the .start_locally() to the HPO example

Tested it locally, with GridSearch and Optuna. Callback function functionality works too.

### DIFF
--- a/clearml/automation/job.py
+++ b/clearml/automation/job.py
@@ -572,7 +572,8 @@ class ClearmlJob(BaseJob):
 
 class LocalClearmlJob(ClearmlJob):
     """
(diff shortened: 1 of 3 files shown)


jkhenning:
@thepycoder LGTM!

PR #526, opened 2021-12-27
Add some convenience functionality to clearml-data

thepycoder:
Added some extra functionality to the `clearml-data` api to support the following use-case:

Say I have a data labeling flow. With as little lines of code as possible, I want to add functionality that quickly creates a new version of my labeled dataset based on the last version, that adds some new labeled samples. Just to keep track of my iterative updates to the dataset, be able to easily share the latest one and  not have to upload all samples every time.

To allow for this I made 3 changes:

1. Create a flag that implicitly creates a dataset and project while running `Dataset.get` if it does not exist yet.

In this way I don't have to either manually create a dataset when I want to start a new labeling process or first check if it exists with a `try catch` block around `Dataset.get` and create a new one on error. Just adding the flag now takes care of that, so I only need 1 line to get or create my dataset.

2. Allow to get a writable copy of the latest dataset by implicitly creating a new one with the latest one as its parent.

Intuitively, this makes a lot of sense. I want to get my dataset, add new files to it and then push the new version up again. This allows me to do just that by creating a new child version.

3. Add automatic upload to the finalize call.

This is quite simply for convenience, less lines. The CLI command will automatically upload all files when finalize is called, so it should be possible to do the same in Python if you want to and are not in need of the extra upload parameters.

None of these changes are breaking current default behaviour, they just aim to add convenience for this type of workflow.

### DIFF
--- a/clearml/backend_interface/util.py
+++ b/clearml/backend_interface/util.py
@@ -52,6 +52,15 @@ def make_message(s, **kwargs):
     return s % args
 
 
+def does_project_exist(session, project_name):
+    res = session.send(projects.GetAllRequest(name=exact_match_regex(project_name), only_fields=['id']))
+    if not res:
+        return None
+    if res.response and res.response.projects:
+        return True
+    return False
+
+
 def get_or_create_project(session, project_name, description=None):
     res = session.send(projects.GetAllRequest(name=exact_match_regex(project_name), only_fields=['id']))
     if not res:
--- a/clearml/backend_interface/util.py
+++ b/clearml/backend_interface/util.py
@@ -52,23 +52,28 @@ def make_message(s, **kwargs):
     return s % args
 
 
-def does_project_exist(session, project_name):
-    res = session.send(projects.GetAllRequest(name=exact_match_regex(project_name), only_fields=['id']))
-    if not res:
-        return None
-    if res.response and res.response.projects:
-        return True
-    return False
-
-
-def get_or_create_project(session, project_name, description=None):
+def get_existing_project(session, project_name):
+    """Return either the project ID if it exists, an empty string if it doesn't or None if backend request failed."""
     res = session.send(projects.GetAllRequest(name=exact_match_regex(project_name), only_fields=['id']))
     if not res:
         return None
     if res.response and res.response.projects:
         return res.response.projects[0].id
-    res = session.send(projects.CreateRequest(name=project_name, description=description or ''))
-    return res.response.id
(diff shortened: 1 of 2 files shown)


clearml/backend_interface/util.py
```diff
@@ -52,6 +52,15 @@ def make_message(s, **kwargs):
     return s % args
 
 
+def does_project_exist(session, project_name):
```
jkhenning, line 55: Wouldn't it make more sense to rename this to `get_existing_project()` (returning `None` if response is falsy, project ID if exists or `""` if no projects are found) and use it in `get_or_create_project()`? less duplicate code 🙂

clearml/datasets/dataset.py
```diff
@@ -463,20 +463,26 @@ def upload(self, show_progress=True, verbose=False, output_url=None, compression
...
-        # type: (bool, bool) -> bool
+    def finalize(self, verbose=False, raise_on_error=True, auto_upload=False):
+        # type: (bool, bool, bool) -> bool
         """
         Finalize the dataset publish dataset Task. upload must first called to verify there are not pending uploads.
         If files do need to be uploaded, it throws an exception (or return False)
 
         :param verbose: If True print verbose progress report
         :param raise_on_error: If True raise exception if dataset finalizing failed
+        :param auto_upload: Automatically upload dataset if not called yet, will upload to default locationz
```
jkhenning, line 474: typo? 🙂

clearml/datasets/dataset.py
```diff
@@ -910,26 +918,37 @@ def get(
         :param dataset_tags: Requested Dataset tags (list of tag strings)
         :param only_completed: Return only if the requested dataset is completed or published
         :param only_published: Return only if the requested dataset is published
+        :param auto_create: Create new dataset if it does not exist yet
+        :param writable_copy: Get a mutable version of the dataset instead, so one can add files to the instance.
```
jkhenning, line 922: I think we should mention here this is a **new** child dataset

clearml/backend_interface/util.py
thepycoder: Agreed! It was embarrassingly DRY... Can you check the newest commit, it should be changed as per your example

clearml/datasets/dataset.py
thepycoder: Is it better now in the new commit?

jkhenning:
@thepycoder LGTM!

PR #533, opened 2022-01-05
Fix and upgrade the SlackMonitor

idantene:
Heya! First PR here -- I can't find any specific code reivew guidelines so I hope this is fine 😉 
This PR contains:
- Fix simple import statements for [Slack](https://slack.dev/python-slack-sdk/v3-migration/index.html#from-slackclient-2-x):
    > The PyPI project is renamed from slackclient to slack_sdk
Importing slack_sdk.* is recommended. You can still use slack.* with deprecation warnings for a while.

- Add the concept of filters to SlackMonitor; essentially a list of callables that accept a `Task` object and return a boolean. Each such filter should decide if the given task should be filtered away (return `True`) or should it not be filtered (return `False`).
- Add arguments for filtering tasks by specific users. This is useful for example to ignore tasks by `Default User` (i.e. a test user), monitoring and remotely executed tasks (that e.g. have a worker ID as username), and more.
- Add `--include_archived` option

### DIFF
--- a/examples/services/monitoring/slack_alerts.py
+++ b/examples/services/monitoring/slack_alerts.py
@@ -55,7 +55,8 @@ class UserFilter:
         all_users = {d["name"]: d["id"] for d in res.json()["data"]["users"]}
         for user in include + exclude:
             if user not in all_users:
-                raise RuntimeError(f"Cannot locate user '{user}' in API!")
+                print(f"Cannot translate user '{user}' to any known user ID - "
+                      f"will use it verbatim")
         self.include = [all_users[user] for user in include]  # Map usernames to user IDs
         self.exclude = [all_users[user] for user in exclude]
 
@@ -277,12 +278,14 @@ def main():
         print("Slack channel was not provided, please run with --channel <channel_name>")
         exit(1)
 
-    # create the user filter
-    user_filter = UserFilter(include=args.include_users, exclude=args.exclude_users)
+    filters = list()
+    # create the user filter if needed
+    if args.include_users or args.exclude_users:
+        filters.append(UserFilter(include=args.include_users, exclude=args.exclude_users))
 
     # create the slack monitoring object
     slack_monitor = SlackMonitor(
-        slack_api_token=args.slack_api, channel=args.channel, message_prefix=args.message_prefix, filters=[user_filter]
+        slack_api_token=args.slack_api, channel=args.channel, message_prefix=args.message_prefix, filters=filters
     )
 
     # configure the monitoring filters
--- a/examples/services/monitoring/slack_alerts.py
+++ b/examples/services/monitoring/slack_alerts.py
@@ -99,6 +99,7 @@ class SlackMonitor(Monitor):
         self.filters = filters or list()
         self.status_alerts = ["failed", ]
         self.include_manual_experiments = False
+        self.include_archived = False
         self._channel_id = None
         self._message_prefix = '{} '.format(message_prefix) if message_prefix else ''
         self.check_credentials()
@@ -158,7 +159,9 @@ class SlackMonitor(Monitor):
 
         :return dict: Example dictionary: {'status': ['failed'], 'order_by': ['-last_update']}
         """
-        filter_tags = ['-archived'] + (['-development'] if not self.include_manual_experiments else [])
+        filter_tags = list() if self.include_archived else ['-archived']
+        if not self.include_manual_experiments:
+            filter_tags.append('-development')
         return dict(status=self.status_alerts, order_by=['-last_update'], system_tags=filter_tags)
 
     def process_task(self, task):
@@ -213,6 +216,7 @@ def main():
                         help='Include experiments running manually (i.e. not by clearml-agent)')
     parser.add_argument('--include_completed_experiments', action="store_true", default=False,
                         help='Include completed experiments (i.e. not just failed experiments)')
+    parser.add_argument('--include_archived', action="store_true", default=False, help='Include archived experiments')
     parser.add_argument('--refresh_rate', type=float, default=10.,
                         help='Set refresh rate of the monitoring service, default every 10.0 sec')
     parser.add_argument('--service_queue', type=str, default='services',
@@ -243,6 +247,7 @@ def main():
     # configure the monitoring filters
     slack_monitor.min_num_iterations = args.min_num_iterations
     slack_monitor.include_manual_experiments = args.include_manual_experiments
+    slack_monitor.include_archived = args.include_archived
     if args.project:
         slack_monitor.set_projects(project_names_re=[args.project])
     if args.include_completed_experiments:
--- a/examples/services/monitoring/slack_alerts.py
+++ b/examples/services/monitoring/slack_alerts.py
@@ -27,7 +27,7 @@ We are done!
 import argparse
 import os
 from time import sleep
-from typing import Optional
+from typing import Optional, List, Callable
 
 from slack_sdk import WebClient
 from slack_sdk.errors import SlackApiError
@@ -41,7 +41,7 @@ class SlackMonitor(Monitor):
     Create a monitoring service that alerts on Task failures / completion in a Slack channel
     """
     def __init__(self, slack_api_token, channel, message_prefix=None, filters=None):
-        # type: (str, str, Optional[str]) -> ()
+        # type: (str, str, Optional[str], Optional[List[Callable[[Task], bool]]]) -> ()
         """
         Create a Slack Monitoring object.
         It will alert on any Task/Experiment that failed or completed
--- a/examples/services/monitoring/slack_alerts.py
+++ b/examples/services/monitoring/slack_alerts.py
@@ -26,8 +26,9 @@ We are done!
 
 import argparse
 import os
+from pathlib import Path
 from time import sleep
-from typing import Optional, List, Callable
+from typing import Optional, Callable, List, Union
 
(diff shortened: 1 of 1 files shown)


examples/services/monitoring/slack_alerts.py
```diff
@@ -27,21 +27,60 @@
...
+        if isinstance(include, str):
+            include = [include]
+        exclude = exclude or list()
+        if isinstance(exclude, str):
+            exclude = [exclude]
+        res = Task._get_default_session().send_request("users", "get_all")
+        if not res.ok:
+            raise RuntimeError("Cannot get list of all users!")
+        all_users = {d['name']: d['id'] for d in res.json()['data']['users']}
+        for user_list in [include, exclude]:
```
jkhenning, line 65: How about:
```python
from itertools import chain
...
for user in chain(include, exclude):
  ...
```

idantene: Yeah good point. I'm a fan of keeping import statements to a minimum when possible -- is this okay then?
```python3
for user in include + exclude:
    ...
```

examples/services/monitoring/slack_alerts.py
```diff
@@ -27,21 +27,60 @@
...
+            for user in user_list:
+                if user not in all_users:
+                    raise RuntimeError(f"Cannot locate user '{user}' in API!")
+        self.include = [all_users[user] for user in include]
```
jkhenning, line 69: Actually, perhaps we should allow user IDs in `include` and `exclude` and only translate the names? So if a value is not found in `all_users`, we simply keep it?

examples/services/monitoring/slack_alerts.py
```diff
@@ -27,21 +27,60 @@
...
+    Either `include` or `exclude` must be passed, but not both.
+    If `include` is given, any task that is not created by any of the users in the 'include' list will be filtered.
+    If `exclude` is given, any task that is created by any of the users in the 'exclude' list will be filtered.
+    If neither is given, this is a no-op.
+    Raises:
+        - ValueError if both include and exclude are given
+        - RuntimeError if the Web API returns an invalid response, or if a given username is not found in the response
+        - KeyError if the Web API return a response of unexpected format
+    """
+    def __init__(self, include=None, exclude=None):
```
jkhenning, line 51: A type hint will be nice 🙂

examples/services/monitoring/slack_alerts.py
```diff
@@ -132,6 +177,9 @@ def process_task(self, task):
             print('Skipping {} experiment id={}, number of iterations {} < {}'.format(
                 task.status, task.id, task.get_last_iteration(), self.min_num_iterations))
             return
+        if any(f(task) for f in self.filters):
+            print("Experiment id={} {} did not pass all filters".format(task.id, task.status))
```
jkhenning, line 181: This seems like a lot to output on a regular basis (if you have many tasks that don't meet the criteria, at least). I would suggest a `verbose` flag (default `False`) that would control this print-out 🙂

examples/services/monitoring/slack_alerts.py
```diff
@@ -168,10 +216,13 @@ def main():
                         help='Include experiments running manually (i.e. not by clearml-agent)')
     parser.add_argument('--include_completed_experiments', action="store_true", default=False,
                         help='Include completed experiments (i.e. not just failed experiments)')
+    parser.add_argument('--include_archived', action="store_true", default=False, help='Include archived experiments')
     parser.add_argument('--refresh_rate', type=float, default=10.,
                         help='Set refresh rate of the monitoring service, default every 10.0 sec')
     parser.add_argument('--service_queue', type=str, default='services',
                         help='Queue name to use when running as a service (default: \'services\'')
+    parser.add_argument('--include_users', type=str, nargs='*', help='Only report tasks from these users')
```
jkhenning, line 224: Since we do not allow both inclusion and exclusion, these should probably go under a mutual exclusive group, something like:
```python
group = parser.add_mutually_exclusive_group()
group.add_argument('--include_users', ...)
group.add_argument('--exclude_users', ...)
...
```

examples/services/monitoring/slack_alerts.py
```diff
@@ -186,13 +237,17 @@ def main():
         print('Slack channel was not provided, please run with --channel <channel_name>')
         exit(1)
 
+    # create the user filter
+    user_filter = UserFilter(include=args.include_users, exclude=args.exclude_users)
```
jkhenning, line 241: Perhaps only create and use the filter if include/exclude are provided?

examples/services/monitoring/slack_alerts.py
jkhenning, line 224: Oh, also - why `nargs='*'` and not `nargs='+'`? I would expect an empty list to be of no effect 🙂

examples/services/monitoring/slack_alerts.py
```diff
@@ -186,13 +237,17 @@ def main():
...
+    
     # create the slack monitoring object
-    slack_monitor = SlackMonitor(
-        slack_api_token=args.slack_api, channel=args.channel, message_prefix=args.message_prefix)
+    slack_monitor = SlackMonitor(slack_api_token=args.slack_api, channel=args.channel,
```
jkhenning, line 244: Last one, I promise! 🙂 - we do prefer to promote Black formatting, so I'd appreciate passing this through a Black formatter 🙂

examples/services/monitoring/slack_alerts.py
jkhenning: Yeah, you can also do:
```python
missing_users = {user for l in (include, exclude) for user in l}.difference(user_list):
if missing_users:
  raise RuntimeError(f"Cannot locate users {', '.join(missing_users)} in API!")
```
😉

examples/services/monitoring/slack_alerts.py
idantene: Is there a way (some regex?) to identify that a given string is an acceptable user ID? I don't mind removing this test (or replacing it with a warning), just that from a WebUI perspective, one would see e.g. `--include_users ['531c83175bn3915nc90', 'jkhenning', '13r7afadji49080a']`, which is maybe confusing (who are these users with those IDs), etc.

examples/services/monitoring/slack_alerts.py
idantene: Hm, the `UserFilter` is anyway a no-op if the `include` and `exclude` are empty -- I thought it's cleaner rather than adding a conditional:

```python3
filters = list()
if args.include_users or args.exclude_users:
    filters.append(UserFilter(...))

slack_monitor = SlackMonitor(..., filters=filters)
```

examples/services/monitoring/slack_alerts.py
idantene: Done! I ran it with line width 120 as agreed -- notice it also formatted the argparse section. Is that okay?

examples/services/monitoring/slack_alerts.py
jkhenning: I know, but adding a filter when nothing needs to be filtered might confuse readers 🙂

ainoam:
@idantene Sure appreciate if once your additions are merged, you can update [the relevant docs](https://github.com/allegroai/clearml-docs/edit/main/docs/guides/services/slack_alerts.md) with the new CLI options.

examples/services/monitoring/slack_alerts.py
jkhenning: Looks great :)

examples/services/monitoring/slack_alerts.py
jkhenning: There isn't really a way, but I think simply leaving strings that can't be matched with user names (with a warning that this is probably an ID) is a very nice solution 🙂

examples/services/monitoring/slack_alerts.py
idantene: Good point :) This is an example after all :D
Updated!

examples/services/monitoring/slack_alerts.py
idantene: Replaced with a warning (in the form of a print :D)

idantene:
> @idantene Sure appreciate if once your additions are merged, you can update [the relevant docs](https://github.com/allegroai/clearml-docs/edit/main/docs/guides/services/slack_alerts.md) with the new CLI options.

Done :) https://github.com/allegroai/clearml-docs/pull/152

examples/services/monitoring/slack_alerts.py
```diff
@@ -26,22 +26,54 @@
...
+            exclude = [exclude]
+        res = Task._get_default_session().send_request("users", "get_all")
+        if not res.ok:
+            raise RuntimeError("Cannot get list of all users!")
+        all_users = {d["name"]: d["id"] for d in res.json()["data"]["users"]}
+        for user in include + exclude:
+            if user not in all_users:
+                print(f"Cannot translate user '{user}' to any known user ID - "
+                      f"will use it verbatim")
+        self.include = [all_users[user] for user in include]  # Map usernames to user IDs
```
jkhenning, line 60: Wouldn't it fail when the user is not in `all_users`? I assume this needs to change to:
```python
self.include = [all_users.get(user, user) for user in include]
self.exclude = [all_users.get(user, user) for user in exclude]
```

idantene: True, forgot about this...

PR #561, opened 2022-01-31
Fix/click

eugen-ajechiloae-clearml:
Fix group arguments in click. Arguments from group commands such as the one seen in `click_multi_cmd.py` would be ignored by ClearML before this patch.

### DIFF
--- a/clearml/binding/click_bind.py
+++ b/clearml/binding/click_bind.py
@@ -15,7 +15,6 @@ class PatchClick:
     _args_type = {}
     _num_commands = 0
     _command_type = 'click.Command'
-    _group_type = 'click.Group'
     _section_name = 'Args'
     _main_task = None
     __remote_task_params = None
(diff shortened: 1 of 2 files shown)


clearml/binding/click_bind.py
```diff
@@ -66,15 +66,15 @@ def _update_task_args(cls):
 
     @staticmethod
     def _command_init(original_fn, self, *args, **kwargs):
-        if self and isinstance(self, Command) and 'name' in kwargs:
-            PatchClick._num_commands += 1
-            if running_remotely():
-                pass
-            else:
+        if self and (isinstance(self, Command) or isinstance(self, Group)) and 'name' in kwargs:
```
jkhenning, line 69: Since `None` has a specific type and `isinstance` allows a shorter syntax, you can do:
`if isinstance(self, (Command, Group)) and 'name' in kwargs:`

eugen-ajechiloae-clearml: done

PR #563, opened 2022-02-02
Fixed requirements in some examples

Rizwan-Hasan:
- Fixed requirements.txt for Reporting examples
- Specified python version for MegEngine in Frameworks example

### DIFF
--- a/examples/reporting/requirements.txt
+++ b/examples/reporting/requirements.txt
@@ -7,4 +7,10 @@ pillow>=4.0
 plotly
 seaborn
 six
+torch
+torchvision
+tqdm
+pytorch-ignite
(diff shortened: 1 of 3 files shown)


examples/reporting/requirements.txt
```diff
@@ -7,4 +7,10 @@ pillow>=4.0
 plotly
 seaborn
 six
+torch
```
jkhenning, line 10: Since the reporting examples should be very simple to use, I suggest not having `torchvision`, `pytorch-ignite` and `tensorboard` here in the requirements and whenever using them, adding a `try/except Import Error` clause that will explain they should be installed

PR #565, opened 2022-02-03
Fixed requirements and refactored code in some examples

Rizwan-Hasan:
- Fixed requirements.txt for Reporting examples
- Specified python version for MegEngine in Frameworks example
- Refactored model_update_pytorch.py example from reporting to frameworks/pytorch


pollfly:
> * Refactored model_update_pytorch.py example from reporting to frameworks/pytorch
Hi @Rizwan-Hasan :) 
Why did you move the example? Originally, I put the example in its current location because I thought the example's main purpose was to demonstrate how to log/report a model using clearml.OutputModel.

jkhenning:
I think that considering the fact the it requires installation of pytorch, it's an acceptable choice

PR #576, opened 2022-02-08
Fix/argparse warning

eugen-ajechiloae-clearml:
When there is an issue patching the Argparser raise a warning for the user (so it won't fail running remotely).

### DIFF
--- a/clearml/binding/args.py
+++ b/clearml/binding/args.py
@@ -37,15 +37,23 @@ class PatchArgumentParser:
     @staticmethod
     def parse_args(self, args=None, namespace=None):
         if PatchArgumentParser._recursion_guard:
-            return {} if not PatchArgumentParser._original_parse_args else \
-                PatchArgumentParser._original_parse_args(self, args=args, namespace=namespace)
+            return (
+                {}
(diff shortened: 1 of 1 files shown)


clearml/binding/args.py
```diff
@@ -36,27 +37,49 @@ def add_subparsers(self, **kwargs):
...
-                PatchArgumentParser._original_parse_args, self, args=args, namespace=namespace)
+                PatchArgumentParser._original_parse_args, self, args=args, namespace=namespace
+            )
+        except Exception as e:
+            result = (
+                {}
+                if not PatchArgumentParser._original_parse_args
+                else PatchArgumentParser._original_parse_args(self, args=args, namespace=namespace)
+            )
+            warnings.warn("Error when patching arguments: %s" % e)
```
jkhenning, line 57: If this is specific to argparse, I think the error should mention argparse, something like "Failed patching argparse arguments: %s"

eugen-ajechiloae-clearml: now having more relevant warnings

shomratalon (approved):

PR #603, opened 2022-03-11
Fix DeprecationWarning triggered by implicit int conversion

vlad-ivanov-name:
In recent versions of Python, there's a warning triggered by `_daemon` function:

```
DeprecationWarning: an integer is required (got type LazyEvalWrapper).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
```

Use explicit conversion to avoid it.

### DIFF
--- a/clearml/backend_interface/task/development/worker.py
+++ b/clearml/backend_interface/task/development/worker.py
@@ -54,10 +54,10 @@ class DevWorker(object):
         last_ping = time()
         while self._task is not None:
             try:
-                if self._exit_event.wait(min(self.ping_period, self.report_period)):
+                if self._exit_event.wait(min(int(self.ping_period), int(self.report_period))):
                     return
                 # send ping request
-                if self._support_ping and (time() - last_ping) >= self.ping_period:
+                if self._support_ping and (time() - last_ping) >= int(self.ping_period):
                     self.ping()
                     last_ping = time()
                 if self._dev_stop_signal:


clearml/backend_interface/task/development/worker.py
```diff
@@ -54,10 +54,10 @@ def _daemon(self):
         last_ping = time()
         while self._task is not None:
             try:
-                if self._exit_event.wait(min(self.ping_period, self.report_period)):
```
jkhenning, line 57: @vlad-ivanov-name I'm not sure I understand - both `self.ping_period` and `self.report_period` are floats numbers, `min` should accept floats, and the timeout argument to `wait()` is float as well - why does a cast to int happen at all?

vlad-ivanov-name: You're right, it should be a float. It seems that this problem was fixed in fcdc561f :) So I'll close this PR.

As for the warning, I'm also not sure why it appears. It's a built-in, so from this native call:

https://github.com/python/cpython/blob/ccbc31ecf3a08ef626be9bbb099f0ce801142fc8/Python/bltinmodule.c#L1575

the following function is called:

https://github.com/python/cpython/blob/ccbc31ecf3a08ef626be9bbb099f0ce801142fc8/Objects/object.c#L710

from there, `tp_richcompare` will be invoked. For the lazy wrapper class it means that one of overridden `__` methods will be invoked, perhaps this int conversion is triggering it?

https://github.com/allegroai/clearml/blob/ac1750b442e4c76ea6c9390f6913ce6da644d52c/clearml/utilities/proxy_object.py#L271 

Either way, it seems like this warning is from kind of "last resort" conversion attempt in `_PyLong_FromNbIndexOrNbInt`

https://github.com/python/cpython/blob/ccbc31ecf3a08ef626be9bbb099f0ce801142fc8/Objects/longobject.c#L242

jkhenning: > It seems that this problem was fixed in https://github.com/allegroai/clearml/commit/fcdc561f5f053312aee9a81b6c8e0561819ed86b :) So I'll close this PR.

Yeah... sorry, just saw this after my answer 🙁 - In any case, thanks for the PR and your time!! 🙂

Issue #614, opened 2022-03-22
Model config serialization fails with Keras Lambda layer on TF 2.5


bzamecnik:
When training a Keras model, upon upgrade to TF 2.5 tracking to ClearML (even with latest 1.3.1) started raising a warning:

```
clearml - WARNING - ('Could not serialize configuration dictionary:\n' ....
```
In particular the config contains a lambda function (K.mean) and ClearML can't serialize it neither to Hocon, nor to JSON. Eg. part of the model's config:

```
{'class_name': 'Lambda',
   'config': {'name': 'row_projection',
    'trainable': True,
    'dtype': 'float32',
    'function': ('4wAAAAAAAAAAAwAAAAwAAAAfAAAAc0QAAAB5CogAfAB8AY4BUwAEAHQAdAFmAmsKcj4BAAEAAQB0\nAogBfAB8AYMDfQJ8AnQDagRrCXI4fAJTAIIAWQBuAlgAZAFTACkCekJDYWxsIHRhcmdldCwgYW5k\nIGZhbGwgYmFjayBvbiBkaXNwYXRjaGVycyBpZiB0aGVyZSBpcyBhIFR5cGVFcnJvci5OKQXaCVR5\ncGVFcnJvctoKVmFsdWVFcnJvctoIZGlzcGF0Y2jaDE9wRGlzcGF0Y2hlctoNTk9UX1NVUFBPUlRF\nRCkD2gRhcmdz2gZrd2FyZ3PaBnJlc3VsdCkC2gZ0YXJnZXTaB3dyYXBwZXKpAPpJL3Vzci9sb2Nh\nbC9saWIvcHl0aG9uMy42L2Rpc3QtcGFja2FnZXMvdGVuc29yZmxvdy9weXRob24vdXRpbC9kaXNw\nYXRjaC5weXIKAAAAywAAAHMOAAAAAAICAQoBEgMMAQoBBAI=\n',
     None,
     (<function tensorflow.python.keras.backend.mean(x, axis=None, keepdims=False)>,
      <function tensorflow.python.keras.backend.mean(x, axis=None, keepdims=False)>)),
    'function_type': 'lambda',
    'module': 'tensorflow.python.keras.backend',
    'output_shape': None,
    'output_shape_type': 'raw',
    'output_shape_module': None,
    'arguments': {'axis': -2}},
   'name': 'row_projection',
   'inbound_nodes': [[['conv2d_7', 0, 0, {}]]]},
```

[...]

PR #616, opened 2022-03-22
WIP: Fix Keras model config serialization in PatchKerasModelIO.

bzamecnik:
Issue: #614

In case the model contains some Lambda layer the Model._updated_config()
function may return objects that are not serializable, such as function
objects (eg. `K.mean`). Its serialization to JSON then fails.

This change adds proper serialization (as done in Model.to_json()) to
two places where the model configuration is passed to OutputModel for
being serialized: _updated_config(), _update_outputmodel().

The return value of patching _updated_config() preserves the
non-serializable objects as some other code may depend on them.

---

I tried this patch in my model and it stopped logging the warning. However, the model config still didn't appear in the ClearML UI.


clearml/binding/frameworks/tensorflow_bind.py
```diff
@@ -1667,21 +1667,26 @@ def _updated_config(original_fn, self):
             return config
 
         try:
+            import json
```
jkhenning, line 1670: Since `json` is part of the standard library, I think we should import it at the top of the file

jkhenning: Also, I think the to make this even more resilient, we should wrap this section in a `try/except` section that falls back to the original config in case of an error, so we'll end up with something like:
```python
try:
    # noinspection PyBroadException
    try:
        from tensorflow.python.util.serialization import get_json_type
        # Model._updated_config() may contain non-serializable objects
        safe_config = json.loads(json.dumps(config, default=get_json_type))
    except Exception:
        safe_config = config

    # there is no actual file, so we create the OutputModel without one
    ...
```

jkhenning:
Hi @bzamecnik, sorry for not getting to this sooner - I've added two comments 🙏

jkhenning:
Hi @bzamecnik, any news on this?

bzamecnik:
Hi @jkhenning, sorry I forgot about the issue. I'll fix it and push again.

clearml/binding/frameworks/tensorflow_bind.py
```diff
@@ -1832,7 +1841,11 @@ def _update_outputmodel(self, *args, **kwargs):
             # this will already generate an output model
             # noinspection PyBroadException
             try:
-                config = self._updated_config()
+                import json
```
eugen-ajechiloae-clearml, line 1844: You imported json at the top, no need to import it here again

clearml/binding/frameworks/tensorflow_bind.py
```diff
@@ -1832,7 +1841,11 @@ def _update_outputmodel(self, *args, **kwargs):
...
+                from tensorflow.python.util.serialization import get_json_type
+                # Model._updated_config() may contain non-serializable objects
+                unsafe_config = self._updated_config()
```
eugen-ajechiloae-clearml, line 1847: I would move this line at the top of the try/except block and have it like this:

```py
try:
    config = {}   # default config val
    config = self._updated_config()   # updated to unsafe config (might fail)
    from tensorflow.python.util.serialization import get_json_type  # might fail
    config = json.loads(json.dumps(unsafe_config, default=get_json_type))   # updated to safe config, might fail
except Exception:
    pass
```

This way we safely build the config from a default value to the safe value you mentioned.

eugen-ajechiloae-clearml:
Hi @bzamecnik! Left 2 small comments

clearml/binding/frameworks/tensorflow_bind.py
bzamecnik: Aah, that's a mistake. :)

clearml/binding/frameworks/tensorflow_bind.py
bzamecnik: This approach seems to be less readable since we overwrite the `config` variable multiple times. Here we need to go through each line and think what the state is when the next lines raises an exception. Also there's likely a bug. In case the last line (json deserialization) fails, `config` will contain the unsafe config. In the MR it would fall back to an empty dict. In contrast, setting config to an empty dict in the except clause is pretty clear.

eugen-ajechiloae-clearml: If falling back to the unsafe config is worse than falling back to an empty one I agree with your comment 👍

Issue #531, opened 2022-01-04
Fix Bool Arg Casting


FawadAhmed322:
When cloning task with args which are bools, it behaves incorrectly. I assume it stores the args as strings thus the bool False is stored as "False" as a string. Only issue is, when the task is cloned or reset and enqueued, it incorrectly casts it giving it a value of True. Since all strings if cast to bool are assigned True.

FawadAhmed322:
the type defined in argparser is nargs="?"

FawadAhmed322:
However I also had that issue when I tried to reproduce this arg type=bool defined in the parser.add_argument

bmartinn:
Hi @FawadAhmed322 
Just making sure we are on the same page here, will the following reproduce the issue?

```
from argparse import ArgumentParser
from clearml import Task

if __name__ == '__main__':
    parser = ArgumentParser()
    parser.add_argument('--str', nargs="?", type=str, default=False)

    args = parser.parse_args()

    task = Task.init(project_name='examples', task_name='argparser')

    print(args)
```

FawadAhmed322:
I'm not sure in this condition since you're passing type=str but passing a boolean so not sure how it'll work. but in the case that you set type to bool, it would be false the first time you run it but will be set to True once you run through ClearML.

bmartinn:
Hi @FawadAhmed322 

> but in the case that you set type to bool, it would be false the first time you run it but will be set to True once you run through ClearML.

Wait, I think I lost you. What exactly was the way you configured the argparser ? (For some reason I was under the impression this was: `nargs="?", type=str, default=False` )

FawadAhmed322:
Hi **@bmartinn** 

in my original code, it was written nargs="?", but no type was defined. Because of it, it set the type to Bool or something. Let me show to you the code.

`parser.add_argument('--resume', nargs='?', default=False, help='resume most recent training')`

This was it.

bmartinn:
Okay, thanks @FawadAhmed322 for clearing it out.
Notice `nargs` will assume `str` as the type unless otherwise specified (see [here](https://docs.python.org/3/library/argparse.html#nargs))

Can I assume the idea is to pass the Task ID to continue training from, as the value of the argument?

For reference reproducible code snippet
```python
from argparse import ArgumentParser
from clearml import Task

if __name__ == '__main__':
    parser = ArgumentParser()
    parser.add_argument('--resume', nargs='?', default=False, help='resume most recent training - pass task ID')

[...]

PR #622, opened 2022-03-25
cast_basic_type says "False" is True

bewt85:
Currently `cast_basic_type("False", "bool") == True`.  This causes issues if you use `task.get_parameters_as_dict(cast=True)`.

This might be related to https://github.com/allegroai/clearml/issues/531

Environment:

* python 3.9
* clearml 1.3.1
* debian

How to reproduce:

```python
from clearml import Task
t = Task.create(project_name="Debugging", task_name="test_get_params")
t.set_parameters_as_dict({"General": {"should_be_false": False}})
t.get_parameters_as_dict(cast=True)["General"]["should_be_false"] # True
```

This PR means that any string except 'False' is True

### DIFF
--- a/clearml/utilities/proxy_object.py
+++ b/clearml/utilities/proxy_object.py
@@ -97,7 +97,8 @@ def cast_basic_type(value, type_str):
     if not type_str:
         return value
 
-    basic_types = {str(getattr(v, '__name__', v)): v for v in (float, int, bool, str, list, tuple, dict)}
+    basic_types = {str(getattr(v, '__name__', v)): v for v in (float, int, str, list, tuple, dict)}
+    basic_types['bool'] = lambda v: False if v == 'False' else True
 
     parts = type_str.split('/')
     # nested = len(parts) > 1


clearml/utilities/proxy_object.py
```diff
@@ -97,7 +97,8 @@ def cast_basic_type(value, type_str):
     if not type_str:
         return value
 
-    basic_types = {str(getattr(v, '__name__', v)): v for v in (float, int, bool, str, list, tuple, dict)}
+    basic_types = {str(getattr(v, '__name__', v)): v for v in (float, int, str, list, tuple, dict)}
+    basic_types['bool'] = lambda v: False if v == 'False' else True
```
jkhenning, line 101: @bewt85 I think we need to support lower-case as well, and raise a `ValueError` in case the value is not as expected (we should also support empty values), something like:
```python
def convert_bool(s):
  s = s.strip().lower()
  if s == "true":
    return True
  elif s == "false" or not s:
    return False
  raise ValueError("Invalid value (boolean literal expected): %s".format(s))

basic_types['bool'] = convert_bool
```

WDYT?

bewt85: That sounds like a great idea.  I'll update the PR.

bewt85:
Thank you!

PR #683, opened 2022-05-29
deprecation warning fixed

Rizwan-Hasan:
Deprecation warning `rename cnn_learner to vision_learner since we now support models other than CNNs too` is fixed.
FastAI introduced this change from v2.6.0, here’s the [CHANGELOG.md](https://github.com/fastai/fastai/blob/master/CHANGELOG.md#260)


jkhenning:
This is new in 2.6.0, isn't it? in that case, we should update the requirements as well...

Rizwan-Hasan:
Thanks for the suggestion. Now requirements are updated.

PR #737, opened 2022-08-04
Update Task.init docstring to include behaviour when executing remotely

mmiller-max:
Following conversation in Slack, I couldn't find where in the documentation it says that `Task.init()` just returns the current task (ignoring any arguments given to it) when being executed remotely by a clearml-agent, so added it to the docstring.

### DIFF
--- a/clearml/task.py
+++ b/clearml/task.py
@@ -234,11 +234,10 @@ class Task(_Task):
         - The Task has run before (the same ``task_name`` and ``project_name``), and (a) it stored models and / or
           artifacts, or (b) its status is Published , or (c) it is Archived.
         - A new Task is forced by calling ``Task.init`` with ``reuse_last_task_id=False``.
-        - ``Task.init`` is called on a clearml-agent (i.e. the task is being executed remotely).
 
         Otherwise, the already initialized Task object for the same ``task_name`` and ``project_name`` is returned,
-        or, when being executed remotely, the current task is returned with any arguments supplied to ``Task.init``
-        being ignored.
+        or, when being executed remotely on a clearml-agent, the current task is returned with any the ``task_name``
+        and ``project_name`` arguments being ignored.
 
         .. note::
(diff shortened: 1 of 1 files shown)


clearml/task.py
```diff
@@ -234,8 +234,11 @@ def init(
         - The Task has run before (the same ``task_name`` and ``project_name``), and (a) it stored models and / or
           artifacts, or (b) its status is Published , or (c) it is Archived.
         - A new Task is forced by calling ``Task.init`` with ``reuse_last_task_id=False``.
+        - ``Task.init`` is called on a clearml-agent (i.e. the task is being executed remotely).
```
jkhenning, line 237: This is not accurate - When executed remotely, the task (experiment) itself already exists

jkhenning: Perhaps the confusion is that this section pertains to the creation of a new task instance in the backend, not to the creation of a new object on the client side representing the server-side task

clearml/task.py
```diff
@@ -234,8 +234,11 @@ def init(
...
 
-        Otherwise, the already initialized Task object for the same ``task_name`` and ``project_name`` is returned.
+        Otherwise, the already initialized Task object for the same ``task_name`` and ``project_name`` is returned,
+        or, when being executed remotely, the current task is returned with any arguments supplied to ``Task.init``
+        being ignored.
```
jkhenning, line 241: Not all arguments are ignored, for example all the `auto_connect...` arguments are always considered

clearml/task.py
mmiller-max: Oh sorry I though the list was the situations where a new task was not created - will adjust

clearml/task.py
mmiller-max: OK ,is it better for it to say that the project and task name arguments are ignored or just leave out any mention of ignoring arguments?

jkhenning: Better just say the object represents the existing task from the backend

Issue #769, opened 2022-09-06
Pipeline that invokes a step function twice will crash if cloned and run from the UI


tonyd:
## Describe the bug
If a pipeline created with decorators invokes the same step function twice, two step tasks are correctly created, the second having "_1" appended to its name. However, if that pipeline is then cloned and run via the UI, execution will fail with a crash.

The problem occurs because the cloned pipeline task has a configuration for the second, "XXX_1" step, but there is no `task_factory_func`. This field is properly filled in for the first "XXX" step because it is established when the pipeline decorators are executed, and the configuration is merged on top of that.

## To reproduce
This test code creates the pipeline by running locally, but the cloned pipeline run needs two workers connected to the "default" queue. This code must be available to the worker via a Git repository.
```
from clearml.automation.controller import PipelineDecorator
from clearml import TaskTypes, Task


@PipelineDecorator.component(return_values=['data_frame'], cache=True, task_type=TaskTypes.data_processing)
def step_one(mock_parameter='mock'):
    print('step_one')
    import pandas as pd
    data_frame = pd.DataFrame()
    return data_frame


@PipelineDecorator.pipeline(name='custom pipeline logic', project='examples', version='0.0.5')
def executing_pipeline(mock_parameter='mock'):
    # Use the pipeline argument to start the pipeline and pass it ot the first step
    print('launch step one')
    data_frame = step_one(mock_parameter)
    data_frame = step_one(mock_parameter + '_2')

    return Task.current_task()


if __name__ == '__main__':
    # Run locally to create a new pipeline instance and have it complete before returning from the function call.
    PipelineDecorator.run_locally()

    task = executing_pipeline()

    from clearml.backend_api.services.v2_20.queues import GetDefaultRequest
    queue_id = task.send(GetDefaultRequest()).response.id

    # Clone and start a new instance of this pipeline; this is the same call the UI makes.
    from clearml.backend_api.services.v2_20.pipelines import StartPipelineRequest
    req = StartPipelineRequest(task=task.id, queue=queue_id)
    resp = task.send(req).response
    new_task_id = resp.pipeline

    print("started new pipeline {}".format(new_task_id))

    # Wait for the new pipeline to execute.
    from clearml.backend_api.services.v2_20.events import GetTaskLogRequest
    import time

    while True:
        task = Task.get_task(task_id=new_task_id)
[...]

Issue #757, opened 2022-08-26
Use of monitor_models on a pipeline component overwrites all earlier output models on the pipeline task


tonyd:
Thank you for helping us making ClearML better!

## Describe the bug
Use of the monitor_models param on a pipeline component to copy a model up to the pipeline task causes any output models already on the pipeline task from an earlier step to be discarded, even if there is no name conflict.

## To reproduce
```
from clearml.automation.controller import PipelineDecorator

@PipelineDecorator.component(monitor_models=['model1'])
def step1():
    from clearml import Task, OutputModel

    model = OutputModel(name="model1")
    with open('model1.xxx', 'w') as f: pass
    model.update_weights(weights_filename='model1.xxx')

@PipelineDecorator.component(monitor_models=['model2'])
def step2():
    from clearml import Task, OutputModel

    model = OutputModel(name="model2")
    with open('model2.xxx', 'w') as f: pass
    model.update_weights(weights_filename='model2.xxx')

@PipelineDecorator.pipeline(name='test pipeline monitor models', project='tests', version='1')
def pipeline():
    from clearml import Task

    print(step1())

    print("after step1:")
    PipelineDecorator.get_current_pipeline().update_execution_plot()
    Task.current_task().reload()
    for m in Task.current_task().models.output:
        print(m.name)

    print(step2())

    print("after step2:")
    PipelineDecorator.get_current_pipeline().update_execution_plot()
    Task.current_task().reload()
    for m in Task.current_task().models.output:
        print(m.name)

if __name__ == '__main__':
    PipelineDecorator.run_locally()
    pipeline()
```

## Expected behaviour
```
ClearML Task: created new task id=8f1f99dce1c34821932fdcbf40670bf5
ClearML results page: https://app.clear.ml/projects/323c182596f34d73934d0b7905eeb81d/experiments/8f1f99dce1c34821932fdcbf40670bf5/output/log
ClearML pipeline page: https://app.clear.ml/pipelines/323c182596f34d73934d0b7905eeb81d/experiments/8f1f99dce1c34821932fdcbf40670bf5
Launching step [step1]
ClearML results page: https://app.clear.ml/projects/323c182596f34d73934d0b7905eeb81d/experiments/42967e99b8f946f6a650972ca1b950a2/output/log
ClearML Monitor: GPU monitoring failed getting GPU reading, switching off GPU monitoring
2022-08-26 12:46:41,857 - clearml.Task - INFO - Waiting to finish uploads
[...]

PR #770, opened 2022-09-06
#769: fix crash when running cloned pipeline that invokes a step twice

tonyd:
## Related Issue \ discussion
https://github.com/allegroai/clearml/issues/757

## Patch Description
Fills in `task_factory_func` for a second or subsequent invocation of the same step function in a cloned pipeline.

This code uses `node.job_code_section` to find the code for the function. This field was only filled in under certain conditions, but the field appears to be unused for anything, so I have changed it to always be filled in when the step is associated with a function.

Once this is fixed, there is a secondary issue that causes the execution of this cloned pipeline to create and execute an extra "XXX_2" task step in place of the loaded-from-configuration "XXX_1" step, which is ignored. This is also fixed in this patch by checking `_launched_step_names`, not `_nodes`, when creating new step names.

## Testing Instructions


clearml/automation/controller.py
```diff
@@ -3079,11 +3084,12 @@ def result_wrapper(a_func_return, return_index):
                     _node.parents = []
                     # find a new name
                     counter = 1
-                    while _node.name in cls._singleton._nodes:
+                    while _node.name in cls._singleton._launched_step_names:
```
eugen-ajechiloae-clearml, line 3087: This looks more like a workaround than a fix. We shouldn't create an extra step at all (or ignore other steps). Any idea why this might happen?

tonyd: There is clearly some confusion about when nodes should be created - the decorator itself creates nodes, and then calling the function tries to match up with that pre-created node, and creates a new node only if it's not the first invocation. A cloned pipeline has pre-existing nodes from being loaded. I did try to make a localized fix to the problem.

Issue #409, opened 2021-07-24
Question about  folder structure when uploading subproject artifacts


tsutsu3:
I use self-hosted clearml and MinIO.

I was able to create a subproject and upload the artifacts with the following settings.

```python
task = Task.init(
    project_name="examples/sub",
    task_name="task_name",
    output_uri="s3://localhost:9000/test",
)
```

As a result, the folder structure of MinIO looks like this:

```
└── examples%2Fsub
    └── task_name.ec4533155df84de49098959ec8ccb6ca
        └── artifacts
```

Is it possible to create a 'sub' folder under 'examples' ?

JDennisJ:
Hi @tsutsu3 ,


So just to verify, this should be the result:
```
└── examples
  └──sub
    └── task_name.ec4533155df84de49098959ec8ccb6ca
       └── artifacts
```
>Is it possible to create a 'sub' folder under 'examples' ?

Currently no, but this should be possible, let me update here once this is possible 😃

JDennisJ:
Hi @tsutsu3 ,

I did some checks here, and actually no it will not be possible and can cause some issues because the subfolder might mix with the parent folder data.

Subfolders can always moved, the naming is just for easier manage when need to (Originally it was the UUID of the project as a folder name).

Can you specify what is the use case for making subfolders? Do you have a specific scenario?

tsutsu3:
> Can you specify what is the use case for making subfolders? Do you have a specific scenario?

Not particularly. Will manage it with flat projects. 😉

bmartinn:
For the time being I'm closing the issue, feel free to reopen ...

kajetanj:
In my case this is causing issues with the object URL
```
https://<some_id>-apidata.googleusercontent.com/download/storage/v1/b/<bucket_name>/o/<project_name>%2F<subproject_name>%2F<experiment_name>.<experiment_id>%2Fartifacts%2F<filename>?jk=<key>
```
when I click it I get
```
No such object: <my_bucket_name>/<project_name>/<subproject_name>/<experiment_name>.<id>/artifacts/<filename>
```
Which is correct, the object is located in
```
<my_bucket_name>/<project_name>%2F<subproject_name>/<experiment_name>.<id>/artifacts/<filename>
```
but the URL doesn't escape `%` sign and `%2F` is interpreted as `/`.
I can't modify the URL, I get `Bad Locked Domain Request`.

jkhenning:
[...]

PR #802, opened 2022-10-18
fix bad artefact URLs when using Google object storage

kajetanj:
## Related Issue \ discussion
https://github.com/allegroai/clearml/issues/409#issuecomment-1282235811

## Testing Instructions
To reproduce the issue run an experiment within a sub-project and upload an artefact when using Google object storage as a backend.


jkhenning:
Hi @kajetanj,

we tested this issue with the latest ClearML SDK version and it seems to work - I think we added a fix for it a while back - can you please re-test with the latest version to verify?

kajetanj:
It does work with version 1.7.2 and uploads to `[bucket]/[project]/[subproject]`. Thanks!

Issue #805, opened 2022-10-24
Freezing on connection break


frolovconst:
## Description
In case of a broken connection to the API server, the function [Task._delete_artifacts()](https://github.com/allegroai/clearml/blob/25df5efe74972624671df2ae97a3c629eb0c5322/clearml/backend_interface/task/task.py#L1360) freezes the main thread. If you try to delete an artifact using the function and the connection happens to be broken at the time, the code enters an [infinite loop](https://github.com/allegroai/clearml/blob/b598ee58666675d5b6ac775f30c682dd3fa6ec62/clearml/backend_interface/base.py#L57). The number of retries settings in the `api.http.retries` config section have no effect. And the [`raise_on_errors=False`](https://github.com/allegroai/clearml/blob/b598ee58666675d5b6ac775f30c682dd3fa6ec62/clearml/backend_interface/task/task.py#L1377) flag prevents the code from throwing an exception and moving on.

## To reproduce
```
task = Task.init(reuse_last_task_id=task_id, continue_last_task=0)
```
_Connection to the API server breaks_
```
delete_artifact(task, 'artifact_name')
task.close()
```

## Expected behaviour
I'm expecting that after a certain number of retries I get informed (e.g., in the form of an exception) about the failed artifact deletion attempt and my ML process can resume.  
To add some context, I need to implement a "keep N best" checkpoints logic. On top of that I need to be able to have different N-numbers for the local and remote storages. See the Slack discussion [here](https://clearml.slack.com/archives/CTK20V944/p1666110774455869).

## Environment
* Server type: self hosted
* ClearML SDK Version 1.7.2
* ClearML Server Version (Only for self hosted): 1.3.0-165
* Python Version 3.9.1
* OS Debian 10
## Related Discussion
[If this continues a slack thread, please provide a link to the original slack thread. ](https://clearml.slack.com/archives/CTK20V944/p1666110774455869)

pollfly:
Hey @frolovconst! [v1.8.0](https://github.com/allegroai/clearml/releases/tag/v1.8.0) is now out with the `raise_on_errors` argument of the `Task.delete_artifacts()` method

PR #806, opened 2022-10-24
'raise_on_errors' flag exposed to the parameters of Task._delete_arti…

frolovconst:
Avoids infinite loop in case of problems with the connection to the API server.

## Related Issue \ discussion
[805](https://github.com/allegroai/clearml/issues/805#issue-1420255351)

## Patch Description
If the flag `raise_on_errors` is set to True, in case of connectivity problems, it's possible to catch the exception and avoid an infinite loop of retries

## Testing Instructions
1. Initiate a task
2. Upload an artifact to it
3. Simulate a connection break
4. Execute `task.delete_artifacts()`
5. Catch the `requests.exceptions.ConnectTimeout`
Result: the main thread doesn't freeze

## Other Information
In order not to use a private function externally, a public version added as well.

### DIFF
--- a/clearml/backend_interface/task/task.py
+++ b/clearml/backend_interface/task/task.py
@@ -1360,12 +1360,12 @@ class Task(IdObjectBase, AccessMixin, SetupUploadMixin):
     def delete_artifacts(self, artifact_names, raise_on_errors):
         # type: (Sequence[str], bool) -> bool
         """
-                Delete a list of artifacts, by artifact name, from the Task.
+        Delete a list of artifacts, by artifact name, from the Task.
 
-                :param list artifact_names: list of artifact names
-                :param bool raise_on_errors: if True, do not suppress connectivity related exceptions
-                :return: True if successful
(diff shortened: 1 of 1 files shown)


clearml/backend_interface/task/task.py
```diff
@@ -1357,12 +1357,24 @@ def _add_artifacts(self, artifacts_list):
                 self._edit(execution=execution)
         return self.data.execution.artifacts or []
 
-    def _delete_artifacts(self, artifact_names):
-        # type: (Sequence[str]) -> bool
+    def delete_artifacts(self, artifact_names, raise_on_errors):
```
jkhenning, line 1360: @frolovconst can you please make `raise_on_errors=True` by default?

jkhenning: updated the comment :)

frolovconst: Yes, sure :)

Issue #852, opened 2022-12-16
Infinite loop in case of a broken connection to the API server


frolovconst:
## Description
_Issue similar to #805_ 
In case of a broken connection to the API server, the functions [Task.set_parameters()](https://github.com/allegroai/clearml/blob/90f30e8d9a5ca9a1afa6b2e5ffccb96b0afe9c78/clearml/backend_interface/task/task.py#L1083) and [Task.upload_artifact()](https://github.com/allegroai/clearml/blob/90f30e8d9a5ca9a1afa6b2e5ffccb96b0afe9c78/clearml/task.py#L1900) freeze the main thread. If you use these functions and the connection happens to be broken at the time, the code enters an [infinite loop](https://github.com/allegroai/clearml/blob/b598ee58666675d5b6ac775f30c682dd3fa6ec62/clearml/backend_interface/base.py#L57). The number of retries settings in the `api.http.retries` config section have no effect. And the [`raise_on_errors=False`](https://github.com/allegroai/clearml/blob/b598ee58666675d5b6ac775f30c682dd3fa6ec62/clearml/backend_interface/task/task.py#L1377) flag prevents the code from throwing an exception and moving on.

## To reproduce
```
task = Task.init(reuse_last_task_id=task_id, continue_last_task=0)
```
_Connection to the API server breaks_
```
task.set_parameter('param_name', 'param_value')
task.close()
```
or 
```
from tempfile import NamedTemporaryFile
with NamedTemporaryFile('wb') as f:
    f.write(b"Your code shall not pass!")
    f.flush()
    task.upload_artifact(name="artifact_name", artifact_object=f.name, wait_on_upload=True)
task.close()
```

## Expected behaviour
I'm expecting that after a certain number of retries I get informed (e.g., in the form of an exception) about the failed attempt to update the task and my ML process can resume.   

## Environment
* Server type: self hosted
* ClearML SDK Version 1.8.3
* ClearML Server Version (Only for self hosted): 1.3.0-165
* Python Version 3.9.1
* OS Debian 10

PR #853, opened 2022-12-16
Exception in case of a connection break propagated to the upper level instead of being suppressed in an infinite loop

frolovconst:
Avoids infinite loop in case of problems with the connection to the API server. The PR is similar to [another one](https://github.com/allegroai/clearml/pull/806) related to connectivity problems

## Related Issue \ discussion
#852 <br/>

## Patch Description
If the flag `raise_on_errors` is set to True, in case of connectivity problems, it's possible to catch the exception and avoid an infinite loop of retries<br/>

## Testing Instructions
1. Initiate a task
2. Simulate a connection break
3. Upload an artifact to the task, or set a parameter to it
4. Catch the `requests.exceptions.ConnectionError`
5. Result: the main thread doesn't freeze 

## Other information  
There's another way to avoid the main thread freezing - make the code which updates tasks asynchronous. Anyway, with an infinite loop taking place there's little sense in stacking task update jobs which will potentially never finish.

### DIFF
--- a/clearml/backend_interface/task/task.py
+++ b/clearml/backend_interface/task/task.py
@@ -1438,7 +1438,7 @@ class Task(IdObjectBase, AccessMixin, SetupUploadMixin):
         with self._edit_lock:
             if Session.check_min_api_version("2.13") and not self._offline_mode:
                 req = tasks.AddOrUpdateArtifactsRequest(task=self.task_id, artifacts=artifacts_list, force=True)
-                res = self.send(req, raise_on_errors=False)
+                res = self.send(req, raise_on_errors=True)
                 if not res or not res.response or not res.response.updated:
                     return None
                 self.reload()
@@ -2317,7 +2317,7 @@ class Task(IdObjectBase, AccessMixin, SetupUploadMixin):
                         "[status={} fields={}]".format(status, list(kwargs.keys()))
                     )
 
-            res = self.send(tasks.EditRequest(task=self.id, force=True, **kwargs), raise_on_errors=False)
+            res = self.send(tasks.EditRequest(task=self.id, force=True, **kwargs), raise_on_errors=True)
             return res
 
     def _update_requirements(self, requirements):


jkhenning:
Hi @frolovconst,

Did you have a chance to test this against a non-responsive server? This kind of change can potentially affect the SDK's ability to wait on servers during downtime. Although it doesn't look like there's an issue, I think it would be prudent to make sure 🙂

frolovconst:
> Hi @frolovconst,
> 
> Did you have a chance to test this against a non-responsive server? This kind of change can potentially affect the SDK's ability to wait on servers during downtime. Although it doesn't look like there's an issue, I think it would be prudent to make sure 🙂

Hi, @jkhenning 
Yes, I did. It's by testing ClearML against a non-responsive server that I found out about this issue. And of course I made sure my solution works when I completed it.
The point is, there is a `api.http.retries.connect` setting in the clearml.conf file which allows you to regulate the time you are willing to wait for a server experiencing downtime, and this is not impacted by the change suggested. Apart from that, currently, the code will always wait for an unresponsive server forever, additionally, hanging the main thread, which makes little sense if you have expectations from your training process to complete within predictable time.

jkhenning:
@frolovconst thanks for the detailed answer 🙂 
Since this is used in some automation contexts, I think the right thing to do is to allow controlling this option using `raise_on_error` bool argument which we can than pass when using in the appropriate places (like from `def _add_artifact`), WDYT?

frolovconst:
@jkhenning Hi, Jake!
Yeah, I see your point. But could you please explain which scenarios you have on your mind and how this argument might be used?
In my opinion, the only impact the value `raise_on_error=False` has now is trigger an infinite loop of retries while the number of retries is already controlled the clearml.conf, and this setting is being ignored now. So this will just create a chain of passing of a redundant argument through several functions.

jkhenning:
I'm not sure I understand what you mean - In case we get a ConnectionError, changing this behaviour will cause `_send()` to raise the connection error - this means any code right now calling a flow that uses the calls changed in this PR will now might not return a value (or not return, depending on implementation), but will raise an error. This includes several places where the functions are described as returning a boolean for success/failure etc, which will both break the documented behavior and might cause issues in places where the code now relies on the fact that these errors are silent.

Issue #825, opened 2022-11-21
Bug: StorageHelper.delete fails when used with Path substitutions


john-zielke-snkeos:
## Describe the bug
The function `clearml.storage.helper.StorageHelper.delete` does not honor the path substitution rules. While all upload/download functions call ```self._canonize_url(remote_path)``` first, delete does not. This leads to an error when uploading files using `Dataset.add_external_files(...)`, as it calls `clearml.storage.helper.StorageHelper.check_write_permissions(...)` which calls `clearml.storage.helper.StorageHelper.delete`. The previous call to `StorageHelper.upload_from_stream(...)` succeeds with the path substitution.

## To reproduce
Not working code, as I cannot reproduce internal paths etc on other setups
```
from clearml import Dataset
from clearml.storage.helper import StorageHelper

StorageHelper.add_path_substitution("s3://minio.server:8000","s3://minio-instance.myintranet.net:8000")

# Preprocessing code here
output_uri = "s3://minio.server:8000/data/clearml"
dataset = Dataset.create(
    dataset_name="test-dataset",
    dataset_project="test-project",
    dataset_version="1.1",
    output_uri=output_uri,
    description="Test dataset",
)

dataset.add_external_files(
            source_url="s3://minio.server:8000/data/my-file.txt",verbose=True
        )
dataset.upload()
dataset.finalize()
```

## Expected behaviour
The path is also rewritten for delete and check access succeeds.

## Environment
* Server type: self hosted
* ClearML SDK Version: 1.7.2
* ClearML Server Version: 1.7.0-232
* Python Version: 3.8.3
* OS: Windows

erezalg:
Hi @john-zielke-snkeos ,

Thanks for investigating this! this makes sense. 
Would you care to open a [PR](https://github.com/allegroai/clearml/blob/master/docs/contributing.md)? we always value community contribution :smile:

john-zielke-snkeos:
Thank you! I will try to do so, just want to make sure that I didn't get something wrong and that it might be intended behavior.

PR #877, opened 2023-01-10
Bugfix for storage with path substitutions

john-zielke-snkeos:
## Related Issue \ discussion
Related to [this previous issue](https://github.com/allegroai/clearml/issues/825), I noticed a few more problems when using path substitutions together with datasets.

## Patch Description
The patch makes both StorageHelper.list() and StorageHelper.exists_file() respect the path substitutions that are set up.

It also makes _get_object_size_bytes only use a special case for the boto driver if the result has the content_length attribute (therefore it is a boto s3 object), as the ListResult that will still occur in combination with the Boto3Driver only has the size attribute.

Finally it handles a problem where the get_metadata of the storage_manager will return the url for the objects with the canonized base url(i.e. with the local path substitutions applied), instead of the original remote_url. There might be a better way to fix this last issue, so suggestions are welcome.

### DIFF
--- a/clearml/storage/helper.py
+++ b/clearml/storage/helper.py
@@ -781,7 +781,7 @@ class StorageHelper(object):
             Listed relative to the base path.
         """
         if prefix:
-            prefix = self._canonize_url(dest_path)
+            prefix = self._canonize_url(prefix)
             if prefix.startswith(self._base_url):
                 prefix = prefix[len(self._base_url):]
                 if self._base_url != "file://":
--- a/clearml/storage/helper.py
+++ b/clearml/storage/helper.py
@@ -619,7 +619,7 @@ class StorageHelper(object):
             if isinstance(self._driver, _HttpDriver) and obj:
                 obj = self._driver._get_download_object(obj)  # noqa
                 size = int(obj.headers.get("Content-Length", 0))
-            elif isinstance(self._driver, _Boto3Driver) and obj:
+            elif isinstance(self._driver, _Boto3Driver) and obj and hasattr(obj, "content_length"):
                 # noinspection PyBroadException
                 try:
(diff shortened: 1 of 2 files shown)


clearml/storage/helper.py
```diff
@@ -781,6 +781,7 @@ def list(self, prefix=None, with_metadata=False):
             Listed relative to the base path.
         """
         if prefix:
+            prefix = self._canonize_url(dest_path)
```
jkhenning, line 784: Aren't you simply overriding the prefix here?

clearml/storage/helper.py
```diff
@@ -1237,6 +1238,7 @@ def close_async_threads():
                 pass
 
     def exists_file(self, remote_url):
+        remote_url = self._canonize_url(remote_url)
```
jkhenning, line 1241: I'm not sure this is right - the path substitutions are meant to replace a remote prefix with a local prefix, but since whet you're getting here is a remote URL and are trying to query the remote storage, applying the substitution will basically guarantee you won't find it... Do you have an example this fixes?

john-zielke-snkeos: But isn't the "storage" pointed to by the path substitution supposed to have the same data as the remote storage. Following example: I have an object storage that is hosted in two locations and synced with each other (strongly consistent). Now my dataset uses one url to point to this storage, and one (or both) locations use a path substitution to map this url to the one that is reachable locally. If the path substitution is not honored here, there is no way for it to check if this storage actually exists.

clearml/storage/helper.py
john-zielke-snkeos: Sorry, this was a copy-paste error, it's fixed now

PR #891, opened 2023-01-17
Remove `exit_process` restriction on `execute_remotely`

mmiller-max:
## Related Issue \ discussion
[Slack discussion](https://clearml.slack.com/archives/CTK20V944/p1673286661348369).

## Patch Description
This change removes a restriction that I don't think is required (would love to know whether this is correct or not!) and blocks my proposed workflow. 

The workflow I want to do is:

1. Create a `task` using `Task.init`
2. Configure the `task`
3. Call `task.execute_remotely()` so that the `task` is reset to `pending` in the server
4. Back to step 1 (without exiting the python process)

### DIFF
--- a/clearml/task.py
+++ b/clearml/task.py
@@ -2477,10 +2477,6 @@ class Task(_Task):
             - ``True`` - Exit the process (exit(0)).
             - ``False`` - Do not exit the process.
 
-            .. warning::
-
-                If ``clone==False``, then ``exit_process`` must be ``True``.
-
         :return Task: return the task object of the newly generated remotely executing task
         """
         # do nothing, we are running remotely
@@ -2497,11 +2493,6 @@ class Task(_Task):
             enqueue_task = Task.clone(source_task=self) if clone else self
             Task.enqueue(task=enqueue_task, queue_name=queue_name)
             return
-
-        if not clone and not exit_process:
-            raise ValueError(
-                "clone==False and exit_process==False is not supported. "
-                "Task enqueuing itself must exit the process afterwards.")
 
         # make sure we analyze the process
         if self.status in (Task.TaskStatusEnum.in_progress,):


eugen-ajechiloae-clearml:
Hi @mmiller-max ! We do not support initing multiple tasks in the same process (we will warn the users about this in a future patch...). I suggest you do the init in different processes instead. Note that you must not call `init` in the parent process. Something like (I didn't test the code, but it should be ok):
```py
from clearml import Task, Logger
from multiprocessing import Process
from tqdm import tqdm


def init_task(iteration):
    t = Task.init(
        project_name="Project",
        task_name="name"  + str(iteration)
    )
    # something here
    t.execute_remotely()


NUMBER_OF_TASKS = 1500
BATCH_SIZE = 16
processes = []
for i in tqdm(range(NUMBER_OF_TASKS)):
    p = Process(target=init_task, args=(i,))
    p.start()
    processes.append(p)
    if len(processes) == BATCH_SIZE:
        for process in processes:
            process.join()
        processes = []
for process in processes:
    process.join()
```

mmiller-max:
Hi @eugen-ajechiloae-clearml , thanks for getting back to me!

> We do not support initing multiple tasks in the same process

It would be helpful to understand why not, is it not supported because of a design decision or because of complexity in code? For example, is there a problem with doing this:

```python
for i in range(0,10):
    task = Task.init(reuse_last_task_id=False)
    # do some work
    task.close()
```

As an extension to the above, you can mimic the behaviour of `execute_remotely(exit_process=False, clone=False)` (with the change in this PR):

```python
for i in range(0,10):
    task = Task.init(reuse_last_task_id=False)
    # do some configuration
    task.close()
    task.reset()  # sets it to draft, can be picked up by a worker
```

I guess the issue with this is that when it is executed on an agent, each agent will iterate through the `for` loop and run the task ten times (as `Task.init` returns the current task on an agent) resetting it each time afterwards.

> By the way, what is your use-case? Maybe we can suggest something else

I'd like to set up several tasks that are launched from one process and then executed on agents. Currently we're doing it by

```python
parent_task = Task.init(...)
for kwargs in kwarg_sets:
    child_task = parent_task.create_function_task(experiment_callable, task_name=**kwargs)
    _launch_agent(child_task)  # in house function that spins up an agent and executes a task
```
But this requires creating a parent task, which doesn't add anything. But it does get around the issue I mention above and  still has all the niceness of `Task.init` (like capturing python environment, uncommitted changes etc).

As I'm explaining what I want I think I now understand why `execute_remotely` functions the way it does! So this PR definitely doesn't make sense. 

It would still be helpful to know if there is an issue with my first example above though, of initting and executing multiple tasks on one machine in one process.

PR #893, opened 2023-01-19
Fix dosctring typo

pollfly:
## Related Issue \ discussion
https://clearml.slack.com/archives/CTK20V944/p1674066273984049

## Patch Description
Dataset.sync_folder return type fixed from (int, int) >>> (int, int, int)


pollfly:
Fixed in different commit

Issue #923, opened 2023-02-17
ClearML 1.9 does not correctly upload debug samples (introduced in 1.9).


jday1:
## Describe the bug

Running an application with clearml 1.9.1 I run into this error: 
```
2023-02-08 15:17:25,538 - clearml - WARNING - Failed decoding debug image [553, 509, 3]
2023-02-08 15:17:25,539 - clearml - WARNING - Error: I/O operation on closed file.
```

And by debug samples are not available in the ClearML UI.

When running with 1.8.3 installed, I do not have this issue. I investigated the git history and found that this particular line of this commit is the culprit:

https://github.com/allegroai/clearml/commit/d723299f1eac75ccb87b21ad802a9683761d3bd5#diff-4808825f4905980a3e4ea49ea3261dde4a6e35c1654557ac2b3a1211cb63659eL402

I have raised a PR with what I believe to be an adequate fix (just reverting to what was previously there). 

## To reproduce
Exact steps to reproduce the bug. Provide example code if possible.

The code is quite nested by I've tried to extract out the important parts (`summary_writer`  is a tensorboard logger).
```
self.figure, (ax1, ax2, axc) = plt.subplots(1, 3, figsize=(total_width, total_height), facecolor="white")

self.summary_writer = self.tb_logger.experiment

self.summary_writer.add_figure(Partition.TRAINING.value, train_plot.figure, global_step=self.current_epoch + 1) 
```
The `train_plot.figure` is a `matplotlib` figure created using seaborn.

## Expected behaviour
What is the expected behaviour? What should've happened but didn't?

I should not receive warning messages and should be able to see debug samples in the ClearML UI 

## Environment
* Server type app.clear.ml
* ClearML SDK Version 1.9.1
* Python Version 3.10.4
* OS (Windows \ Linux \ Macos) Have checked and this occurs on all OS versions.
## Related Discussion
If this continues a slack thread, please provide a link to the original slack thread.

https://clearml.slack.com/archives/CTK20V944/p1675869748638919

jkhenning:
Hi @jday1 , thanks! we'll take a look at this soon!

vfdev-5:
I see the same issue with 1.9.0, 1.9.1 and 1.9.2rc2 versions. Please could you make new release 1.9.2rc3 version with the fix included. Thanks!

PR #924, opened 2023-02-17
PR relating to bug #923.

jday1:
## Related Issue \ discussion
#923 

## Patch Description

## Testing Instructions

### DIFF
--- a/clearml/binding/frameworks/tensorflow_bind.py
+++ b/clearml/binding/frameworks/tensorflow_bind.py
@@ -8,16 +8,15 @@ from functools import partial
 from io import BytesIO
 from mimetypes import guess_extension
 from tempfile import mkstemp
-from typing import Any, Union
 
 import numpy as np
 import six
 from PIL import Image
 
-from ...debugging.log import LoggerRoot
 from ..frameworks import _patched_call, WeightsFileHandler, _Empty
 from ..import_bind import PostImportHookPatching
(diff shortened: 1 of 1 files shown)


clearml/binding/frameworks/tensorflow_bind.py
```diff
@@ -8,16 +8,15 @@
...
 
 import numpy as np
 import six
 from PIL import Image
 
-from ...debugging.log import LoggerRoot
 from ..frameworks import _patched_call, WeightsFileHandler, _Empty
 from ..import_bind import PostImportHookPatching
 from ...config import running_remotely
+from ...debugging.log import LoggerRoot
```
eugen-ajechiloae-clearml, line 19: Please don't change the imports at all. `Any` and `Union` need to be imported, because we have some type annotations that require them.

eugen-ajechiloae-clearml (changes requested):
Thanks for the PR. Left 1 comment

clearml/binding/frameworks/tensorflow_bind.py
jday1: Apologies. I was ensuring that the file conformed to the pull requests section of the contributing guide (https://github.com/allegroai/clearml/blob/master/docs/contributing.md?plain=1#L56).  

I have reverted the change as requested.

PR #927, opened 2023-02-20
Improve the hyper_parameter example

Make42:
I added more examples and made sure the name of the file reflects the contents.
In fact, the example file did already contain more than hyper parameter logging, before my changes: In the web app stuffs that are logged via connect_configuration is under "configuration objects", not "hyperparameters".

- Introduce sections
- Add example for user properties
- Add an example for hyperparameters with TaskParameters
- Rename file to reflect the naming in the web app

This PR relates to https://github.com/allegroai/clearml-docs/issues/486.

### DIFF
--- a/clearml/task.py
+++ b/clearml/task.py
@@ -1728,9 +1728,7 @@ class Task(_Task):
         Enables you to manually shut down the task from the process which opened the task.
 
         This method does not terminate the (current) Python process, in contrast to :meth:`Task.mark_completed`.
-        Note, that if :meth:`Task.mark_completed` is called from a different process it will not terminate
-        the process from which :meth:`Task.mark_completed` was called, but the process which created the task.
-        
+
         After having :meth:`Task.close` -d a task, the respective object cannot be used anymore and
         methods like :meth:`Task.connect` or :meth:`Task.connect_configuration` will throw a `ValueError`.
         In order to obtain an object representing the task again, use methods like :meth:`Task.get_task`.
--- a/clearml/task.py
+++ b/clearml/task.py
@@ -1725,10 +1725,10 @@ class Task(_Task):
     def close(self):
         """
         Closes the current Task and changes its status to "Completed".
-        Enables you to manually shutdown the task.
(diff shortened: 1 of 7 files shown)


Make42:
I corrected docstrings. Because my other pull request was not included yet, it seems they are lumped together.

clearml/task.py
```diff
@@ -1725,10 +1725,13 @@ def reset(self, set_started_on_success=False, force=False):
     def close(self):
         """
         Closes the current Task and changes its status to "Completed".
-        Enables you to manually shutdown the task.
-        This method does not terminate the current Python process, in contrast to :meth:`Task.mark_completed`.
+        Enables you to manually shut down the task from the process which opened the task.
+
+        This method does not terminate the (current) Python process, in contrast to :meth:`Task.mark_completed`.
+        Note, that if :meth:`Task.mark_completed` is called from a different process it will not terminate
```
alex-burlacu-clear-ml, line 1731: Given that this sentence (line 1731-1732) describes in detail the behavior of `Task.mark_completed` and not of `Task.close` so I would suggest it would be better placed in the `Task.mark_completed`'s docstring.

Make42: I included this information into the docstring for mark_completed as well, as I am sure you have seen. I included this information here as well as to ensure that there are no misunderstandings.

Make42: @alex-burlacu-clear-ml - do you have an argument against my reasoning? If yes, which?

jkhenning: @Make42 in general we do not like documentation duplication. I believe leaving line 1730 with a reference to `Task.mark_completed` is a good addition and helps users connect the two behaviors, but repeating the other method's docstring is excessive and might also lead to inaccuracies later on when the code evolves. Please remove lines 1731-1732 and we can merge it so it will be in the next upcoming release.

Make42: @jkhenning , @alex-burlacu-clear-ml : Done 👍. Please merge :-).

Issue #963, opened 2023-03-29
`get_reported_scalars` is incorrectly documented for `x_axis` value `timestamp`. 


jday1:
## Describe the bug

`get_reported_scalars` is incorrectly documented for `x_axis` value `timestamp`. 

## To reproduce
Exact steps to reproduce the bug. Provide example code if possible.

Follow the source code.

`get_reported_scalars` -> `ScalarMetricsIterHistogramRequest`

```
    :param key: Histogram x axis to use: iter - iteration number iso_time - event
        time as ISO formatted string timestamp - event timestamp as milliseconds since
        epoch
```

## Expected behaviour
What is the expected behaviour? What should've happened but didn't?

Documentation should be correct.

## Environment
* Server type (self hosted \ app.clear.ml)
* ClearML SDK Version
* ClearML Server Version (Only for self hosted). Can be found on the bottom right corner of the settings screen.
* Python Version
* OS (Windows \ Linux \ Macos)
## Related Discussion
If this continues a slack thread, please provide a link to the original slack thread.

jkhenning:
@jday1 I'm closing the issue. Please reopen if required.

PR #964, opened 2023-03-29
#963: Modified documentation to reflect what actually happens

jday1:
## Related Issue \ discussion

See https://github.com/allegroai/clearml/issues/963

## Patch Description

## Testing Instructions


clearml/backend_interface/task/task.py
```diff
@@ -1948,7 +1948,7 @@ def get_reported_scalars(
         :param int max_samples: Maximum samples per series to return. Default is 0 returning all scalars.
             With sample limit, average scalar values inside sampling window.
         :param str x_axis: scalar x_axis, possible values:
-            'iter': iteration (default), 'timestamp': seconds from start, 'iso_time': absolute time
+            'iter': iteration (default), 'timestamp': absolute time as epoch milliseconds, 'iso_time': absolute time
```
jkhenning, line 1951: Perhaps stick to the server API's definition of `timestamp as milliseconds since epoch`?

jday1: Done

PR #972, opened 2023-04-09
Create pylint.yml

bai2020-05:
## Related Issue \ discussion

## Patch Description

## Testing Instructions

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,23 @@
+name: Pylint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10"]
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


jkhenning:
@bai2020-05 can you please provide some context for this PR?

jkhenning:
@bai2020-05 , I think that since we internally use Flake8, we should start with it - WDYT?

john-zielke-snkeos:
Yes, Flake8 is definitely a good idea, but does it provide auto-formatting as well? I have used Flake8 for the linting, but not for the code formatting, but if it provides that as well, it would be enough

jkhenning:
@bai2020-05 is this still relevant?

john-zielke-snkeos:
I am not sure it is relevant for the author, but it is still relevant in general IMO

jkhenning:
@john-zielke-snkeos Since we use Flake8 internally (btw I think its much more common for open source projects), I think we should have a new PR using it, WDYT?

john-zielke-snkeos:
Yes, it probably makes sense if you just add your config/setup for this to the repo so other contributors can easily use that.

PR #1000, opened 2023-04-27
AWS driver fixes

achaiah:
## Related Issue \ discussion
N/A

## Patch Description
- Fixes AWS credential access that uses a token - existing code does not take into consideration AWS credentials that use key/secret/token triplet to access resources.

- Fixes instantiation within VPC without AvailabilityZones. In these situations you must supply a SubnetID instead of AvailabilityZone.

- Extends EBS volume creation to include existing volumes (specified by `ebs_snapshot_id`). Note: In current code implementation EBS is attached but never mounted.

## Testing Instructions
Credential authentication fails when attempting to use key/secret/token (due to the obvious lack of ability to specify a token)

SubnetID is a required parameter in certain VPC configurations where Availability Zones are not configured.

### DIFF
--- a/clearml/automation/aws_driver.py
+++ b/clearml/automation/aws_driver.py
@@ -20,12 +20,12 @@ except ImportError as err:
         "install with: pip install boto3"
     ) from err
 
-
 @attr.s
 class AWSDriver(CloudDriver):
     """AWS Driver"""
     aws_access_key_id = attr.ib(validator=instance_of(str), default='')
     aws_secret_access_key = attr.ib(validator=instance_of(str), default='')
+    aws_session_token = attr.ib(validator=instance_of(str), default='')
     aws_region = attr.ib(validator=instance_of(str), default='')
     use_credentials_chain = attr.ib(validator=instance_of(bool), default=False)
     use_iam_instance_profile = attr.ib(validator=instance_of(bool), default=False)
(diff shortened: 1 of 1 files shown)


clearml/automation/aws_driver.py
```diff
@@ -47,29 +48,49 @@ def from_config(cls, config):
...
+                    }
+                }
+            ]
+            
+        if resource_conf.get("subnet_id", None):
+            launch_specification["SubnetId"] = resource_conf["subnet_id"]
+        elif resource_conf.get("availability_zone", None):
+            launch_specification["Placement"] = {"AvailabilityZone": resource_conf["availability_zone"]}
+        else:
+            raise Exception('subnet_id or availability_zone must to be specified in the config')
```
jkhenning, line 93: Hi @achaiah, I'm not sure this is entirely correct, can you point me to the documentation for this?

jkhenning:
@achaiah is this still relevant?

achaiah:
Yes, this is still definitely an issue unless it has been fixed separately from this PR.

jkhenning:
OK, do you have any reply for the question I raised in the code comment?

achaiah:
I did reply to your question above. At this moment I don't have official AWS documentation beyond what the AWS API error message requested me to supply (i.e. SubnetId is required if no default AvailabilityZone is configured).

jkhenning:
I'm sorry, I didn't see the reply...

PR #1019, opened 2023-05-21
fix config parsing for the case of empty string keys

radugrosu:
## Patch Description

Solves this bug: 
```python
from clearml.utilities.pyhocon.config_parser import ConfigFactory
ConfigFactory.from_dict({"": 1})
```
fails with `IndexError: list index out of range`.


jkhenning:
Hi @radugrosu,

As however this is part of the tokenizer, I'm worried there might be side-effects - since this is part of pyhocon, and the code was not changed there, I suggest contributing this there, and than, once they approve it, we can do it here as well.

WDYT?

PR #1031, opened 2023-06-06
Handle web app on standard HTTP and HTTPS port

pktiuk:
Closes: https://github.com/allegroai/clearml-server/issues/198

It would handle configurations when web app port is changed to a standard port for HTTP or HTTPS.

It allows configurations like:

    web_server: http://clearml.xyz.local
    api_server: http://clearml.xyz.local:8008
    file_server: http://clearml.xyz.local:8081/


pktiuk:
@jkhenning Could you look?

Issue #1076, opened 2023-07-18
Dataset.list_dataset() partial match doesn't work with speical charcters


natephysics:
## Describe the bug
Dataset.list_dataset() partial match option doesn't escape special characters, which breaks matching in many cases. Tracing it back, Task._query_tasks() task_name doesn't work with special characters either. I plan to offer a PR with the fix at the _query_tasks() level unless there's a compelling reason otherwise. 

## To reproduce
List datasets with a task name with characters that need to be escaped normally. An example, `[ABC] Name` won't match even if the task name is identical. 

## Expected behaviour
Without any mention of needing to escape special characters, it might make more sense to do it by default.

jkhenning:
@natephysics docstrings were updated to reflect this, as discussed in #1077. I'm closing this 🙂

PR #1077, opened 2023-07-18
Fix dataset.list dataset() partial match fix for special chars

natephysics:
## Related Issue \ discussion
https://github.com/allegroai/clearml/issues/1076

## Patch Description
This patch fixes Task._query_task() by escaping the task_name attribute passed to the function. Without this, task_names passed with special characters would break the search. 

## Testing Instructions
See: https://github.com/allegroai/clearml/issues/1076

### DIFF
--- a/clearml/task.py
+++ b/clearml/task.py
@@ -9,8 +9,10 @@ import threading
 import time
 from argparse import ArgumentParser
 from logging import getLogger
+from re import escape
 from tempfile import mkstemp, mkdtemp
 from zipfile import ZipFile, ZIP_DEFLATED
+
 
 try:
     # noinspection PyCompatibility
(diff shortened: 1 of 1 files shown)


jkhenning:
Hi @natephysics, methods using `Task._query_tasks()` such as `Task.get_tasks()` and `Task.query_tasks()` actually depend on the string not being escaped since they explicitly state that the `task_name` can be a regex... Perhaps the right way to address it is to add a note to their docstring, reminding users that in case they do not with to use a regex, they should escape the string?

natephysics:
That makes sense. I actually asked in the discussion if it should be addressed at that level. This was mainly to address the issue of it not being clear that Dataset.list_dataset() partial name matching required escaping as well. Is this also one you want to preserve regex formatting for? If so it would be good to document this as it was a point of confusion.

natephysics:
I'll close it.

PR #1117, opened 2023-09-15
Remove `subdir` path from GCP bucket url

materight:
## Patch Description
When running a cloned task remotely and using GS as storage, calling `task.logger.set_default_upload_destination` usually fails with an error like:
```
GET https://storage.googleapis.com/storage/v1/b/clearml/project_a/.pipelines/Main/task_1.d7691550c955409db4622fd74496f51c/artifacts/result_TRAIN_config/result_TRAIN_config.pkl/iam/testPermissions?permissions=storage.objects.get&permissions=storage.objects.update&prettyPrint=false: Not Found
```
This is because the permissions are being tested on a single file instead of a GS bucket. Removing the `subdir` from the bucket url should fix it.

## Testing Instructions
- Run a pipeline with a single task that calls `task.logger.set_default_upload_destination`
- Clone the task and enqueue it again for execution

### DIFF
--- a/clearml/storage/helper.py
+++ b/clearml/storage/helper.py
@@ -894,7 +894,7 @@ class _GoogleCloudStorageDriver(_Driver):
         obj.download_to_filename(str(p))
 
     def test_upload(self, test_path, config, **_):
-        bucket_url = str(furl(scheme=self.scheme, netloc=config.bucket, path=config.subdir))
+        bucket_url = str(furl(scheme=self.scheme, netloc=config.bucket))
         bucket = self.get_container(container_name=bucket_url, config=config).bucket
 
         test_obj = bucket


jkhenning:
Hi @materight, 

If the bucket has a specific subdir defined in the configuration, wouldn't it make sense to check these permissions on the subdir? Or are you saying this is not supported in Google Storage?

materight:
Hi @jkhenning. The second one, this method works only on buckets: https://cloud.google.com/storage/docs/json_api/v1/buckets/testIamPermissions

jkhenning:
That makes sense, but it wouldn't yield the same error 😕

materight:
Sorry, what do you mean?

materight:
I have this in the config file:
```
sdk {
    google.storage {
        project: "myproject"
        credentials_json: ${GOOGLE_APPLICATION_CREDENTIALS}
        pool_connections: 512
        pool_maxsize: 1024
    }
    development {
        default_output_uri: "gs://cerrion-clearml"
    }
}
```

Also in the agent's logs I only see this:
```
sdk.google.storage.project = myproject
sdk.google.storage.credentials_json = /service_account_key.json
sdk.google.storage.pool_connections = 512
sdk.google.storage.pool_maxsize = 1024
sdk.development.default_output_uri = gs://cerrion-clearml
```

Btw I'm using pipelines, the `result_TRAIN_config.pkl` artifact it tries to access is a parameter of the pipeline component.

jkhenning:
And can you reproduce that error message you shared?

materight:
Yes, if I clone the failed task and re-enqueue it I get the same error. But if I run a new task it doesn't happen all the time.

jkhenning:
The config file you attached is from your own workstation? If so, what's the config file used by the agent?

materight:
It's exactly the same for both

materight:
Sure, here: [task_ff8c8917072641bca89f27ddd488d74c.log](https://github.com/allegroai/clearml/files/12675631/task_ff8c8917072641bca89f27ddd488d74c.log)

materight:
Hi @jkhenning any update on this? Would it be possible to merge it and dig in more if the issue reappear?

PR #1128, opened 2023-10-04
address DeprecationWarnings from jsonschema

xquyvu:
## Related Issue \ discussion
N/A<br/>

## Patch Description
Address the DeprecationWarning raised from `jsonschema`. My VSCode testing configuration treats DeprecationWarnings as Error, which means I cannot run any tests with VSCode testing when having clearml installed.

```
E   DeprecationWarning: Importing FormatError directly from the jsonschema package is deprecated and will become an ImportError. Import it from jsonschema.exceptions instead.
```

## Testing Instructions

```python
>>> import jsonschema
>>> jsonschema.FormatError
<stdin>:1: DeprecationWarning: Importing FormatError directly from the jsonschema package is deprecated and will become an ImportError. Import it from jsonschema.exceptions instead.
<class 'jsonschema.exceptions.FormatError'>
```

### DIFF
--- a/clearml/backend_api/session/request.py
+++ b/clearml/backend_api/session/request.py
@@ -2,12 +2,17 @@ import abc
 
 import six
 from jsonschema.exceptions import FormatError, SchemaError, ValidationError
-from referencing.exceptions import Unresolvable
+
+try:
+    # Since `referencing`` only supports Python >= 3.8, this try-except blocks maintain support
+    # for earlier python versions.
+    from referencing.exceptions import Unresolvable
+except ImportError:
+    from jsonschema.exceptions import RefResolutionError as Unresolvable
 
 from .apimodel import ApiModel
 from .datamodel import DataModel
 from .defs import ENV_API_DEFAULT_REQ_METHOD
-
 
 if ENV_API_DEFAULT_REQ_METHOD.exists() and ENV_API_DEFAULT_REQ_METHOD.get().upper() not in ("GET", "POST", "PUT"):
     raise ValueError(
--- a/clearml/backend_api/session/request.py
+++ b/clearml/backend_api/session/request.py
@@ -1,7 +1,8 @@
 import abc
 
-import jsonschema
 import six
+from jsonschema.exceptions import FormatError, SchemaError, ValidationError
+from referencing.exceptions import Unresolvable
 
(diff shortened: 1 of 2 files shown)


clearml/backend_api/session/request.py
```diff
@@ -1,7 +1,8 @@
 import abc
 
-import jsonschema
 import six
+from jsonschema.exceptions import FormatError, SchemaError, ValidationError
+from referencing.exceptions import Unresolvable
```
jkhenning, line 5: @xquyvu referencing is a separate dependency and this import fails since it's not in the SDK's requirements.
I see you basically replaced `RefResolutionError` with `Unresolvable`, however `jsonschema.exception.RefResolutionError` is still available... Can you please use that instead?

jkhenning: Will this also work with older jsonschema versions?

xquyvu: Sorry I don't quite understand, could you elaborate? Are you asking if `referencing` works with older `jsonschema` versions?

jkhenning: Yes, exactly

xquyvu: Hmm I'm not sure how I can test that. How would you do it? I've created a new environment, installed `referencing` after `dev-requirements.txt` and there's no error at import time.

xquyvu: I see. So maybe we'll just silence the deprecation error and pin `jsonschema` to <4.18? Dropping support for python versions may affect a whole user segment (Although I don't know how big it is)

jkhenning: Perhaps try import from referencing as follows:
```python
try:
  from referencing.exceptions import Unresolvable
except ImportError:
  from jsonschema.exceptions import RefResolutionError as Unresolvable
```

And add this to the requirements:
```
referencing<0.40 ; python_version >= '3.8'
```

xquyvu: [use referencing for Python >= 3.8](https://github.com/allegroai/clearml/pull/1128/commits/6d90bdb1bde7c49dd3c144c7a7384405e9a22914)

Thanks! Let me know if there's anything else.

Issue #1177, opened 2023-12-20
Providing a task ID to clearml-data get results in unintuitive behaviour.


harry-optimised:
## Describe the bug
Using the local CLI tool, if you pass an ID of a task rather than a dataset, the result is that the task is moved into the .datasets folder. 

## To reproduce
Locate any task on ClearML. Locally run `clearml-data get --id <task_id>` observe that the task "dissapears", it is moved into a .datasets folder in it's current project.

## Expected behaviour
The task should not be affected, the CLI tool should emit a warning that a task ID was passed not a dataset.

## Environment
Ubuntu 22.04
ClearML SDK version 1.13
Python3.10

jkhenning:
Hi @harry-optimised,

Thanks for this report, we'll fix it as soon as possible 👍

PR #1178, opened 2023-12-20
Fix issue 1177

harry-optimised:
## Related Issue \ discussion
Relates to #1177 - Providing a task ID to clearml-data get results in unintuitive behaviour.

## Patch Description
Bug in original conversation. This patch checks for existance of the dataset tag before proceeding to get.

### DIFF
--- a/clearml/cli/data/__main__.py
+++ b/clearml/cli/data/__main__.py
@@ -9,6 +9,7 @@ from pathlib2 import Path
 
 import clearml.backend_api.session
 from clearml.datasets import Dataset
+from clearml.task import Task
 from clearml.version import __version__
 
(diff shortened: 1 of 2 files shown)


eugen-ajechiloae-clearml:
Hi @harry-optimised ! I don't think that this change should be localized to the CLI. `Dataset.get` should not allow you to get "datasets" that are not tagged appropriately at all. You could modify this function in the SDK to check for the tag instead: https://github.com/allegroai/clearml/blob/5ad082d2e7d740623f1449665541466c0de0331a/clearml/datasets/dataset.py#L1677

jkhenning:
@harry-optimised any update on this?

harry-optimised:
Sorry for delay - agreed, was a bit lazy. I'll close this for now!

Issue #1211, opened 2024-02-23
PR: Add support to TaskScheduler() to accept a function called at runtime to return a task_id


natephysics:
## Proposal Summary
I'm working on a PR that adds new functionality to the task scheduler by adding a new parameter `task_id_function` to `TaskScheduler.add_task()` that takes a callable which has an expected return of a `task_id`. This `task_id_function` function is run at runtime (when the task scheduler would normally execute the scheduled task) and uses the `task_id` returned by the function + the other parameters from `.add_task()` as the scheduled task.

## Motivation
Why is this useful: there's a host of reasons but the biggest one: it gives users much more control over the tasks that are run by the task scheduler. Currently, as far as I can tell, if I wanted to run the most recent task (at runtime) from a given project with a specific tag, it's not possible to do with the task scheduler. I can use the `schedule_function` parameter and create a function that finds and runs the task but then I lose the core advantages of `.add_task()`, no way to specify queues, task_parameters, and task_overrides. Naturally, I could wrap all of that into the function called by `task_parameters` but then I'm basically just writing my own scheduler at that point. This will also let you do some preprocessing before returning the task_id, for example, if you wanted to clean up old tasks. 

## Related Discussion
https://clearml.slack.com/archives/CTK20V944/p1708447659999379?thread_ts=1708445057.172119&cid=CTK20V944

PR #1212, opened 2024-02-23
feat(scheduler): :sparkles: Adds task_id_function to TaskScheduler

natephysics:
## Related Issue \ discussion
https://github.com/allegroai/clearml/issues/1211
https://clearml.slack.com/archives/CTK20V944/p1708447659999379?thread_ts=1708445057.172119&cid=CTK20V944

## Patch Description
Adds new functionality to the task scheduler by adding a new parameter `task_id_function` to `TaskScheduler.add_task()` that takes a callable that has an expected return of a `task_id`. This `task_id_function` function is run at runtime (when the task scheduler would normally execute the scheduled task) and uses the `task_id` returned by the function + the other parameters from `.add_task()` as the scheduled task.

### Motivation
Why is this useful: there's a host of reasons but the biggest one: it gives users much more control over the tasks that are run by the task scheduler. Currently, as far as I can tell, if I wanted to run the most recent task (at runtime) from a given project with a specific tag, it's not possible to do with the task scheduler. I can use the schedule_function parameter and create a function that finds and runs the task but then I lose the core advantages of `.add_task()`, no way to specify queues, `task_parameters`, and `task_overrides`. Naturally, I could wrap all of that into the function called by task_parameters but then I'm basically just writing my own scheduler at that point. This will also let you do some preprocessing before returning the `task_id`, for example, if you wanted to clean up old tasks.

## Testing Instructions
Define a function that returns a `task_id` by whatever means:

```
def yourfunction():
    return 'some_task_id_string'
```

With the task scheduler call `.add_task(..., schedule_task_id=None, task_id_function=yourfunction, ...)`. At the time the task was scheduled to execute, the function should run and the output `task_id` will be passed as if it had originally passed to `schedule_task_id`.

### DIFF
--- a/clearml/automation/scheduler.py
+++ b/clearml/automation/scheduler.py
@@ -19,6 +19,7 @@ class BaseScheduleJob(object):
     name = attrib(type=str, default=None)
     base_task_id = attrib(type=str, default=None)
     base_function = attrib(type=Callable, default=None)
+    base_task_id_function = attrib(type=Callable, default=None)
     queue = attrib(type=str, default=None)
     target_project = attrib(type=str, default=None)
     single_instance = attrib(type=bool, default=False)
@@ -43,9 +44,13 @@ class BaseScheduleJob(object):
         if self.base_function and not self.name:
             raise ValueError("Entry 'name' must be supplied for function scheduling")
         if self.base_task_id and not self.queue:
-            raise ValueError("Target 'queue' must be provided for function scheduling")
-        if not self.base_function and not self.base_task_id:
-            raise ValueError("Either schedule function or task-id must be provided")
+            raise ValueError("Target 'queue' must be provided for task scheduling")
+        if self.base_task_id_function and not self.queue:
+            raise ValueError("Target 'queue' must be provided for task-id function scheduling")
+        if not self.base_function and not self.base_task_id and not self.base_task_id_function:
+            raise ValueError("Either schedule function, task-id, or task-id function must be provided")
+        if self.base_task_id and self.base_task_id_function:
+            raise ValueError("Only one of task-id or task-id function must be provided")
 
     def get_last_executed_task_id(self):
         # type: () -> Optional[str]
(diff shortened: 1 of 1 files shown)


jkhenning:
Thanks @natephysics, we'll take a look at it soon!

clearml/automation/scheduler.py
```diff
@@ -60,11 +65,12 @@ def run(self, task_id):
             self._executed_instances.append(str(task_id))
 
     def get_resolved_target_project(self):
-        if not self.base_task_id or not self.target_project:
+        if not self.base_task_id or not self.target_project or not self.base_task_id_function:
```
eugen-ajechiloae-clearml, line 68: As we either have `base_task_id` or `base_task_id_function`, this will always evaluate to true. The condition should be: `if (not self.base_task_id and not self.base_task_id_function) or not self.target_project:`.

clearml/automation/scheduler.py
```diff
@@ -76,6 +82,19 @@ def get_resolved_target_project(self):
             pass
         return self.target_project
 
+    def get_task_id(self):
+        # type: () -> str
+        if self.base_task_id_function:
+            # validate retrevial of a valid task id
+            try:
+                base_task_id = self.base_task_id_function()
```
eugen-ajechiloae-clearml, line 90: We should cache the result of `self.base_task_id_function` such that it is not called twice.
This is because `self.get_task_id` is called in this function and it is also called before running the job itself. 
If the `self.base_task_id_function` has side-effects, such as cloning a task, then these effects will happen twice, so one might end up with multiple clones for example.

clearml/automation/scheduler.py
```diff
@@ -634,14 +659,15 @@ def add_task(
...
-        task_id = schedule_task_id.id if isinstance(schedule_task_id, Task) else str(schedule_task_id or "")
+        mutually_exclusive(schedule_function=schedule_function, schedule_task_id=schedule_task_id, task_id_function=task_id_function)
+        task_id = schedule_task_id.id if isinstance(schedule_task_id, Task) else str(schedule_task_id or '')
 
         # noinspection PyProtectedMember
         job = ScheduleJob(
             name=name or task_id,
             base_task_id=task_id,
             base_function=schedule_function,
+            base_task_iÍd_function=task_id_function,
```
eugen-ajechiloae-clearml, line 670: Typo here. `base_task_iÍd_function` should be `base_task_id_function`

eugen-ajechiloae-clearml (changes requested):
Thank you for the contribution. I left a few comments.

jkhenning:
@natephysics any update?

natephysics:
Hi Jake.

Yes. I started implementing the suggested changes but the point about not running the function multiple times was something I overlooked. I fixed most of the issues and in theory the function should only run once but when I test it out it appears to run twice still. I tried to figure out exactly why this is the case but I ran out of time. I still plan to finish it but I have to finish up another project first.

ainoam:
@natephysics Should we keep this PR open?

natephysics:
Yes, sorry for the hanging PR.

Issue #1313, opened 2024-08-14
Relative path is incorrect when using `Dataset.add_external_files` for local file without the `file://` prefix


geometrikal:
## Describe the bug

`Dataset.add_external_file` will allow adding external files from a source url without the `file://` prefix, e.g. `/path/to/my/file.csv`. It will create the link correctly in the metadat.json by prepending `file://` however it will remove characters from the file name when creating the relative link.

## To reproduce

```
dataset = Dataset.create(
        dataset_project="tes",
        dataset_name="test"
)
dataset.add_external_files("/mnt/f/0123456789.txt")
dataset.finalize(auto_upload=True)
print(dataset.list_files())
# ['789.txt']
```

## Expected behaviour

Either throw an exception or correctly create the relative path

## Environment
* Server type self hosted
* ClearML SDK Version 1.16.3
* ClearML Server Version 1.16.1-499 • API: 2.30
* Python Version 3.10
* OS  Linux

eugen-ajechiloae-clearml:
Hi @geometrikal ! Thanks for reporting this. We will get back to you once this is fixed.

eugen-ajechiloae-clearml:
Hi @d-vignesh ! Yes, all contributions are welcome, and no-one picked this one up as far as I'm aware.
It looks like when adding a file using add_external_files, the file name will be stripped of the first `len("file://")` characters. I I believe that we don't really support paths that are not prefixed by `file://` at all, so, in this function: https://github.com/allegroai/clearml/blob/b1e39e68973182d66b43edc502b58d955b5bc779/clearml/datasets/dataset.py#L3322
I would add a check: if the file is a local path without `file://` then get the absolute path and add the `file://` prefix.
This should be tested using absolute paths, relative paths, on both Windows and Linux.

d-vignesh:
Hi @eugen-ajechiloae-clearml can be please clarify on below doubt,
when a source_url without `file://` prefix is provided, 
1. the [code](https://github.com/allegroai/clearml/blob/b1e39e68973182d66b43edc502b58d955b5bc779/clearml/storage/helper.py#L2903) defaults to `file://` as base url and hence the metadata is having the expected name value
2. during [constructing](https://github.com/allegroai/clearml/blob/b1e39e68973182d66b43edc502b58d955b5bc779/clearml/datasets/dataset.py#L3368) the relative path
```
link = remote_object.get("name")
relative_path = link[len(source_url):]
if not relative_path:
[...]

PR #1326, opened 2024-09-15
fix for local file uploads without scheme

d-vignesh:
Fix issue on scenario where local file path is provided without scheme in add_external_file functionality.

## Related Issue \ discussion
Fix for [#1313 ](https://github.com/allegroai/clearml/issues/1313)

## Patch Description
If the file added has a local path, the absolute path is fetched and the 'file://' scheme is added.

## Testing Instructions
Described in [#1313]

## Other Information
Tested file addition using absolute path, relative path and remote path. File addition and upload is working fine.
I could only test for Windows system.
image

### DIFF
--- a/clearml/storage/helper.py
+++ b/clearml/storage/helper.py
@@ -3067,6 +3067,14 @@ class StorageHelper(object):
         return self._driver.exists_file(
             container_name=self._container.name if self._container else "", object_name=object_name
         )
+    
+    @classmethod
+    def sanitize_url(cls, remote_url):
+        base_url = cls._resolve_base_url(remote_url)
+        if base_url != 'file://':
+            return remote_url
+        absoulte_path = os.path.abspath(remote_url)
+        return base_url + absoulte_path
 
 
 def normalize_local_path(local_path):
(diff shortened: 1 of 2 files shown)


clearml/storage/helper.py
```diff
@@ -3067,6 +3067,14 @@ def exists_file(self, remote_url):
...
             container_name=self._container.name if self._container else "", object_name=object_name
         )
+    
+    @classmethod
+    def sanitize_url(cls, remote_url):
+        base_url = cls._resolve_base_url(remote_url)
+        if base_url != 'file://':
+            return remote_url
+        absoulte_path = os.path.abspath(remote_url)
+        return base_url + absoulte_path
```
eugen-ajechiloae-clearml, line 3077: Note that, in case a user adds a file that is already prefixed with `file://`, the `os.path.abspath(remote_url)` will return an unexpected value. For example:
```
>>> os.path.abspath("file:///home/ajecc/work")
'/home/ajecc/file:/home/ajecc/work'
```
I suggest checking if the remote_url already starts with `file://`, in which case you don't  need to add the `base_url`

d-vignesh: Thanks for getting it @eugen-ajechiloae-clearml . Sorry i missed this case. Updated the code and tested it.
image

d-vignesh: Hi @eugen-ajechiloae-clearml , please review this change once when you get the time.

PR #1333, opened 2024-09-25
Adding appropriate error when inputing URL's in add_files function (instead of using add_external_files)

morzahavi:


clearml/datasets/dataset.py
```diff
@@ -380,58 +380,75 @@ def add_tags(self, tags):
         :param tags: A list of tags which describe the Task to add.
         """
         self._task.add_tags(tags)
+import re
```
jkhenning, line 383: Please move to the top of the file, and make sure to sort the imports

PR #1349, opened 2024-11-19
Fix #1348

### LINKED ISSUE #1348: Bug: PermissionError: [WinError 5] while uploading dataset
## Describe the Bug
When uploading a dataset using `clearml-data sync` on Windows, the following error intermittently occurs: `PermissionError: [WinError 5] Access is denied`.

## Investigation
The error is triggered on the following line in the code:  
https://github.com/allegroai/clearml/blob/master/clearml/binding/artifacts.py#L1141C18-L1141C19.

This issue is unstable and difficult to reproduce consistently. However, it becomes a significant problem when working with large datasets. For instance, my team faced this error while attempting to upload a dataset containing 700 000 items. The process took approximately 5 hours to compute all file hashes, only for the `WinError 5` to unexpectedly appear.

## Steps to Reproduce
The following script simulates the aforementioned line of code and demonstrates the issue:

```python
import os
from tempfile import mkstemp
from tqdm import trange, tqdm
from shutil import move

for i in trange(100000):
    fd, local_filename = mkstemp()
    os.close(fd)
    fd, temp_filename = mkstemp()
    os.close(fd)
    
    try:
        os.replace(local_filename, temp_filename)
    except PermissionError:
        tqdm.write("PermissionError")
    
    os.remove(temp_filename)
```

When running this script on multiple Windows machines, I observed about 6 `PermissionError` exceptions over 100 000 iterations.  
**Running the script as administrator does not resolve the issue.**

## Possible Solution
I suggest replacing the use of `os.replace(local_filename, temp_filename)` with `shutil.move(local_filename, temp_filename)`. Both methods achieve the same functionality, but `shutil.move` does not raise the `PermissionError` on Windows.

Octoslav:
## Related Issue \ discussion
Fix #1348 

`os.replace` in bundle with `mkstemp` could raise `PermissionError: [WinError 5]` on windows systems.

## Patch Description
Replaced `os.replace` with `shutils.move`.

## Other Information
See details in the issue

### DIFF
--- a/clearml/binding/artifacts.py
+++ b/clearml/binding/artifacts.py
@@ -257,6 +257,7 @@ class Artifacts(object):
     max_preview_size_bytes = 65536
 
     _flush_frequency_sec = 300.
+    _max_tmp_file_replace_attemps = 3
     # notice these two should match
     _save_format = '.csv.gz'
     _compression = 'gzip'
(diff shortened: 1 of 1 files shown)


jkhenning:
@Octoslav, since `os.replace` should be atomic in Linux, and `shutil.move` can have different behaviors (such as copying between devices), can you please make this change specific to windows only?

Octoslav:
@jkhenning Maybe in this case it is better to just wrap `os.replace` in a `try\except` block? Like:

```python
try:
    os.replace(local_filename, temp_filename)
except PermissionError:
    os.replace(local_filename, temp_filename)  # Try again, and if it fails, throw an exception.
```


What do you think?

jkhenning:
If that addresses the issue, I would be fine with it 🙂

Octoslav:
Hi @jkhenning
please, check mr

jkhenning:
Hi @Octoslav, that looks OK, I will merge it

PR #1387, opened 2025-03-10
cleaned up exception handeling in cleanup_service.py

PixelWelt:
## Related Issue \ discussion
Related issue: https://github.com/clearml/clearml/issues/1386
## Patch Description
made the exception handling clearer to understand, it now excepts a ValueError and checks if the deletion was successful

### DIFF
--- a/examples/services/cleanup/cleanup_service.py
+++ b/examples/services/cleanup/cleanup_service.py
@@ -86,10 +86,16 @@ while True:
                 ):
                     logging.info(f'successfully deleted task with id {task.id}')
                 else:
-                    logging.warning(f'could not delete task with id {task.id}, task.delete() returned false')
+                    logging.warning(f'could not delete task with id {task.id}')
             except ValueError as e:
                 logging.warning(
                     "Could not find Task ID={}, {}".format(
+                        task.id, e.message if hasattr(e, "message") else e
+                    )
+                )
+            except Exception as e:
+                logging.warning(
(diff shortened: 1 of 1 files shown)


examples/services/cleanup/cleanup_service.py
```diff
@@ -79,15 +79,18 @@
...
+                if deleted_task.delete(
                     delete_artifacts_and_models=True,
                     skip_models_used_by_other_tasks=True,
                     raise_on_error=False
-                )
-            except Exception as ex:
+                ):
+                    logging.info(f'successfully deleted task with id {task.id}')
+                else:
+                    logging.warning(f'could not delete task with id {task.id}, task.delete() returned false')
```
jkhenning, line 89: I would remove the comment regarding `task.delete()` - it's implicit in the code

examples/services/cleanup/cleanup_service.py
```diff
@@ -79,15 +79,18 @@
             # noinspection PyBroadException
             try:
                 deleted_task = Task.get_task(task_id=task.id)
-                deleted_task.delete(
+                if deleted_task.delete(
                     delete_artifacts_and_models=True,
                     skip_models_used_by_other_tasks=True,
                     raise_on_error=False
-                )
-            except Exception as ex:
```
jkhenning, line 87: Please include a section after ValueError for intercepting Exception and logging a generic error with the contents - in edge cases, it's important to understand the context

Issue #1405, opened 2025-04-22
`StorageManager` affects root logger of `logging`


artium-d:
## Describe the bug

I've recently started using the `StroageManager` as a utility in my algorithm development code (not using `Task` or anything else from `clearml` package right now). 

Lately I've noticed something strange:
I use the builtin Python logging module like so: `logging.info(msg)` (i.e. I use the root logger).

I configure it to have two handlers:
Console (specifically `from rich.logging import RichHandler`)
File (`logging.FileHandler`)

Now, initially everything works as expected, messages get logged to both console and file, but after the first call to `StorageManager().get_local_copy()` the messages are only logged to console and not to the file.

I set the environment variable: `CLEARML_LOG_LEVEL=INFO`
I also checked and before and after the` StorageManager().get_local_copy()` call, the root logger's `level` and `propagate` attributes are not changed.
Also tried calling `flush()` on the file handler, did not help.


## To reproduce

```python

import logging

from rich.logging import RichHandler
from dotenv import load_dotenv
from clearml import StorageManager


load_dotenv()

handlers = [RichHandler(rich_tracebacks=True), logging.FileHandler('log.log', mode='w')]
logging.basicConfig(level=logging.INFO, format="%(message)s", handlers=handlers, force=True)

logging.info('before download')
StorageManager().get_local_copy(remote_path, 'local.path')
logging.info('after download')
```

In the console I get both messages:
```
[04/16/25 09:11:30] INFO     before download                                                                                                                                                                                                                                  reproduce.py:14
[04/16/25 09:11:31] INFO     after download                                                                                                                                                                                                                                   reproduce.py:16
```

In the `log.log` file I only get:

```
before download
```

`remote_path` is in the form "s3://bucket/dir/file.ext"
`load_dotenv()` is for loading the AWS credentials and I also have those variables set:
```
CLEARML_CACHE_DIR=<path to persistent local directory>
CLEARML_SUPPRESS_UPDATE_MESSAGE="1"
CLEARML_LOG_LEVEL="INFO"
```

## Expected behaviour
I would expect the messages to be logged to both console and file.

## Environment
* Server type  - Just SDK, no server
* ClearML SDK Version - 1.18.0
* ClearML Server Version -  Just SDK, no server
* Python Version - 3.12.3
* OS (Windows \ Linux \ Macos) - Linux
## Related Discussion
Link to slack [thread](https://clearml.slack.com/archives/CTK20V944/p1744701946793769)

[...]

PR #1419, opened 2025-05-20
Fix: StorageManager no longer interferes with root logger (#1405)

josephvincent24:
Summary
This PR fixes an issue where using StorageManager.get_local_copy() (or any ClearML utility that triggers logging initialization) would close user-defined logging handlers, such as [logging.FileHandler], if the user already configured logging. This caused log messages to stop appearing in user log files after the first ClearML call.

Details
Root cause:
ClearML’s logging initialization called [logging.config.dictConfig], which resets and closes all existing logging handlers, including those set up by the user.
Fix:
The patch adds a check in clearml/backend_config/log.py so that [dictConfig] is only called if the root logger has no handlers. This prevents ClearML from interfering with user logging setups.
I also added a fix that creates a shallow copy in clearml/debugging/log.py in the clear_logger_handlers() function. This function closes the handler before it is removed and modifies the loop to iterate over a shallow copy of the handlers list.
Test:
Added a test (test_storage_manager.py) to ensure that user-defined handlers remain open and receive log messages after calling StorageManager.get_local_copy().
Impact:
Backwards compatible:
If the user does not configure logging, ClearML will still set up logging as before.
Best practice:
This approach follows standard open-source library practices for interacting with Python’s logging system.
Closes: StorageManager affects root logger of logging #1405 (https://github.com/clearml/clearml/issues/1405)

### DIFF
--- a/clearml/backend_config/log.py
+++ b/clearml/backend_config/log.py
@@ -41,5 +41,9 @@ def initialize(
         Logger.manager.loggerClass = _Logger
 
     if logging_config is not None:
-        # Use deepcopy since Python's logging infrastructure might modify the dict
-        logging.config.dictConfig(deepcopy(dict(logging_config)))
+        root_logger = logging.getLogger()
+        #added a check to avoid calling dictConfig
+        #if the root user has already set up their own logging.
+        #prevents ClearML from closing their handlers.
+        if not root_logger.handlers:
+            logging.config.dictConfig(deepcopy(dict(logging_config)))
--- a/clearml/debugging/log.py
+++ b/clearml/debugging/log.py
@@ -164,8 +164,10 @@ class LoggerRoot(object):
         loggers = [logging.getLogger()] + list(logging.Logger.manager.loggerDict.values())
         for logger in loggers:
             handlers = getattr(logger, "handlers", [])
-            for handler in handlers:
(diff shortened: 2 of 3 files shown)


jkhenning:
@josephvincent24 how would that work? This means that if the user has any logger with a handler already configured, ClearML will simply not add it's default log settings (which by default are setting the ClearML log level, and sets a proper log level to multiple dependency packages).
The main issue, it seems to me, it implicitness - if this change is added, ClearML's logging behavior will change based on whether some code called by the user prior to calling ClearML defined a logger - this is too implicit and breaks backwards compatibility...
Our default logging configuration does specify `disable_existing_loggers: false` which means any existing non-root loggers are not affected.
A simple solution might be some environment variable that controls this new behavior, but I'm trying to think if there's a smarter way to go about it...

josephvincent24:
@jkhenning I also considered this, but I'm not sure it's safe. When dictConfig is called, it closes all existing handlers attached to the root logger, so simply re-adding them will not resume logging as expected because those handler objects are now closed and unusable. Re-initializing them as new handlers will lose some user customization. Also, manipulating handlers in this way might introduce thread safety issues.

jkhenning:
If that's the case, we should at least support a configuration option / env var to control this behavior and keep the default as the old behavior - WDYT?

PR #1454, opened 2025-08-08
Specify Task.TaskTypes as a type using typing.TypeAlias.

jday1:
Allows mypy to correctly identify it as a type.

## Related Issue \ discussion

## Patch Description

Example:

Current Behaviour:

```
from clearml import Task

class Example:
    task_type: Task.TaskTypes = Task.TaskTypes.application
```

```
mypy example.py
example.py:4: error: Variable "clearml.task.Task.TaskTypes" is not valid as a type  [valid-type]
example.py:4: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
Found 1 error in 1 file (checked 1 source file)
```

with fix:
```
mypy example.py             
Success: no issues found in 1 source file
```

## Testing Instructions

See above snippets.
Using `mypy==1.61.1`, `clearml==2.0.2`

### DIFF
--- a/clearml/task.py
+++ b/clearml/task.py
@@ -9,6 +9,7 @@ import time
 from argparse import ArgumentParser
 from logging import getLogger
 from tempfile import mkstemp, mkdtemp
+from typing import TypeAlias
 from zipfile import ZipFile, ZIP_DEFLATED
 
 try:
@@ -161,7 +162,7 @@ class Task(_Task):
         Task. Creating a sub-task always creates a new Task with a new  Task ID.
     """
 
-    TaskTypes = _Task.TaskTypes
+    TaskTypes = _Task.TaskTypes  # type: TypeAlias
 
     NotSet = object()
 


jkhenning:
Hi @jday1, first of all thanks for this contribution.
Two concerns:

1. TypeAlias was added in Python 3.10, so using it without a type guard will break backwards compatibility to older python versions
2. TypeAlias was deprecated since Python 3.12:
_Deprecated since version 3.12: [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is deprecated in favor of the [type](https://docs.python.org/3/reference/simple_stmts.html#type) statement, which creates instances of [TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) and which natively supports forward references. Note that while [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) and [TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) serve similar purposes and have similar names, they are distinct and the latter is not the type of the former. Removal of [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAlias) is not currently planned, but users are encouraged to migrate to [type](https://docs.python.org/3/reference/simple_stmts.html#type) statements._

patrickdasilva-clearml:
Thank you for the pull request @jday1. 

I am closing this PR due to the last comment and lack of activity for many months. If you wish to renew discussion on the topic, either open an issue or speak with our community on [Slack](https://clearml.slack.com).

Issue #1484, opened 2025-10-23
S3 virtual host style support


khab-el:
## Proposal Summary
Support virtual hosted style

## Motivation
My S3 is configured to only accept requests in [virtual hosted style](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html) notation.

patrickdasilva-clearml:
Hi @khab-el, I understand why you would like ClearML to support this feature. Thanks for adding the link to the AWS documentation. We'll discuss internally and get back to you in this issue thread!

PR #1485, opened 2025-10-23
s3-vhs-support: add addressing style support

khab-el:
## Related Issue \ discussion
#1484


jkhenning:
Hi @khab-el , as the default addressing style is `auto`, I suggest not touching it unless specified, and doing something like:
```python
**({"s3": {"addressing_style": _Boto3Driver._addressing_style}} if _Boto3Driver._addressing_style else {}),
```

while also setting the default to `None`

I would also suggest adding that to the documentation under `/docs/clearml.conf` and in `sdk.conf`

khab-el:
@jkhenning thanks for review, fixed

PR #1489, opened 2025-11-09
Pipeline scalability

tomerroditi:
## Patch Description
1. adding many steps (hundreds) is very slow due to repeated pipeline graph figure build
2. how to handle forced stoped tasks was hard coded, made it configurable
3. `pipeline.get_parameters()` does not retrieve updates hyper-parameters from UI changes before calling `pipeline.start()` which reduces the flexibility of changing the pipeline with UI overrides via the pipelines "Run" page.

## Testing Instructions
1. a simple for loop adding many steps (more than 500 for example) to a pipeline (and execute it remotely)
2. set `aborted_nonresponsive_as_running=True` and force abort (non responsive) an executed task, make sure it is reenqueued properly.
3. make a pipeline task, rerun with UI changes, in the used script call `pipeline.get_parameters()` before starting the pipeline and validate values matching the UI changes.

code for tests of 1,3
```
from clearml import PipelineController


def main() -> None:
    """
    Main script to set up and execute the pipeline.
    """
    main_config = {
        "some_params": {
            "param1": 10,
            "param2": "value",
        },
    }

    print("\n\nCreating Pipeline\n\n")
    pipeline = PipelineController(
        name="test",
        project="test",
    )
    
    print("\n\nAdding parameters to Pipeline\n\n")
    pipeline.add_parameter(name="param1", default_value=main_config["some_params"]["param1"])
    pipeline.add_parameter(name="param2", default_value=main_config["some_params"]["param2"])
    
    pipe_params = pipeline.get_parameters()
    print(f"Pipeline parameters (should match UI changes): {pipe_params}\n\n")
    
    print("Adding steps to Pipeline\n\n")
    for i in range(500):
        pipeline.add_step(
            name=f"step_{i}",
            base_task_project="test",
            base_task_name="test_task",
            # update_execution_plot=False # solves overhead when having many steps
        )
    
    pipeline.start(wait=True)


if __name__ == "__main__":
    main()

```

### DIFF
--- a/clearml/automation/controller.py
+++ b/clearml/automation/controller.py
@@ -523,7 +523,7 @@ class PipelineController(object):
         output_uri: Optional[Union[str, bool]] = None,
         continue_behaviour: Optional[dict] = None,
         stage: Optional[str] = None,
-        update_execution_plot: bool = False
+        update_execution_plot: bool = True
     ) -> bool:
         """
         Add a step to the pipeline execution DAG.


eugen-ajechiloae-clearml:
Hi @tomerroditi ! Thank you for the RP! 
We would like to maintain backwards compatibility with the previous behaviour of how we treated abort non responsive tasks as runnnig. That is, setting the aborted_nonresponsive_as_running flag to True by default.
Do you think you could make this change?

tomerroditi:
@eugen-ajechiloae-clearml thanks for reviewing, missed that, updated to True to preserve default behavior.

PR #1509, opened 2026-01-17
Fix Deprecation warnings from pyparsing

hindicator:
## Related Issue \ discussion
https://github.com/clearml/clearml/issues/1506

## Patch Description

## Testing Instructions


jkhenning:
Hi @hindicator,
Thanks for the contribution! Please note that as the camel-case changes were introduced with `pyparsing` 3.0.0, this change also requires changing the `clearml` requirements to require `pyparsing>=3.0`, and it must be verified as working with Python 3.6 (and up) as we still support that Python version.

mscheltienne:
@1505 is more complete- any chance to get this PR moving?

PR #1515, opened 2026-01-26
fix #1514

### LINKED ISSUE #1514: Scheduler is broken: error when subtracting timezone-aware and timezone-naive datetimes
## Describe the bug
The scheduler starts successfully, but scheduled tasks are not executed.
The scheduler log contains the following message:

`Warning: Exception caught during scheduling step: can't subtract offset-naive and offset-aware datetimes`

## To reproduce
Run the scheduler, for example, with the following code:
`from clearml import Task
from clearml.automation import TaskScheduler

def run_test():
    print('Task executed successfully')

if __name__ == '__main__':
    Task.force_store_standalone_script()

    scheduler = TaskScheduler(
        sync_frequency_minutes=15,
        force_create_task_name='Test Scheduler'
    )

    scheduler.add_task(
        schedule_function=run_test,
        name='Test task',
        single_instance=True,
        execute_immediately=True,
        minute=15
    )

    scheduler.start_remotely()`

Make sure that the ClearML SDK version is the latest one, for example 2.1.2.

The task will not be executed, and a warning will appear in the scheduler logs.

## Expected behaviour
The task should be executed every 15 minutes.

## Environment
Server type: app.clear.ml
ClearML SDK Version: 2.1.2
ClearML Server Version: N/A (not self-hosted)
Python Version: 3.12
OS: Linux

## Related Discussion

svolsky:
## Related Issue \ discussion

## Patch Description

## Testing Instructions

### DIFF
--- a/clearml/automation/scheduler.py
+++ b/clearml/automation/scheduler.py
@@ -158,6 +158,7 @@ class ScheduleJob(BaseScheduleJob):
                 day=int(self.day or 1),
                 hour=int(self.hour or 0),
                 minute=int(self.minute or 0),
+                tzinfo=timezone.utc
             )
             if self.weekdays:
                 self._next_run += relativedelta(weekday=self.get_weekday_ord(self.weekdays[0]))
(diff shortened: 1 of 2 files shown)


patrickdasilva-clearml:
Thanks @svolsky for opening a PR. We have published a fix in `clearml==2.1.5`, so we believe this PR is no longer necessary. If you still experience issue #1514 , feel free to re-open the issue.

PR #1521, opened 2026-02-16
Fix syntax error on python 3.14

mscheltienne:
A `return` in a `finally` block has been deprecated since Python 3.12 and now raises a `SyntaxError` in python 3.14.


patrickdasilva-clearml:
We tested on our end and it seems like you caught a bug for Python 3.14+. Thank you @mscheltienne  for your contribution!

Issue #1522, opened 2026-02-17
`os.fork` in a multi-threaded process is deprecated


mscheltienne:
This line: https://github.com/clearml/clearml/blob/bda93b0a41cbcffd6fe7393f20a516d795a219f1/clearml/utilities/process/mp.py#L662

Raises often a `DeprecationWarning` in CIs (since python 3.12): https://docs.python.org/3.13/whatsnew/3.12.html#deprecated

> On POSIX platforms, [os.fork()](https://docs.python.org/3.13/library/os.html#os.fork) can now raise a [DeprecationWarning](https://docs.python.org/3.13/library/exceptions.html#DeprecationWarning) when it can detect being called from a multithreaded process. There has always been a fundamental incompatibility with the POSIX platform when doing so. Even if such code appeared to work. We added the warning to raise awareness as issues encountered by code doing this are becoming more frequent. See the [os.fork()](https://docs.python.org/3.13/library/os.html#os.fork) documentation for more details along with [this discussion on fork being incompatible with threads](https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555) for why we’re now surfacing this longstanding platform compatibility problem to developers.

Overall- it just seems very unsafe and should be replaced with a `spawn` or `forkserver`- forking in a multi-threaded context copies locks in their current state and can lead to deadlocks in the child process.

Edition-X:
# Issue Fixed by PR #1557

This issue has been fixed by **PR #1557** which addresses both this issue and #1556.

## Summary

PR #1557 replaces `os.fork()` with `spawn`-based multiprocessing, which resolves:
- ✅ **This issue (#1522)**: DeprecationWarning and thread safety
- ✅ **Issue #1556**: Memory leak from fork capturing pre-allocated memory

## The Fix

**Before:**
```python
# Line 662 - mp.py
BackgroundMonitor._main_process = os.fork()  # ← Deprecated, unsafe
```

**After:**
```python
# Use spawn context instead
ctx = get_context("spawn")
_main_process = ctx.Process(
    target=cls._background_process_start,
    args=(task_obj_id, cls._sub_process_started, os.getpid()),
)
```

## Why This Fixes the Issue

### 1. No More `os.fork()` Deprecation Warning
- Switched to `multiprocessing.Process` with `spawn` context
- No direct `os.fork()` calls
- Python 3.12+ deprecation warnings eliminated

### 2. Thread-Safe
[...]

PR #1557, opened 2026-03-14
Fix memory leak in background subprocess (issue #1556)

### LINKED ISSUE #1556: Memory leak inside Task.init
## Describe the bug
A memory leak occurs when you allocate resources before calling `Task.init()` and then attempt to free them.

The leak occurs because Task.init() creates a child process via the `fork` call with all of the allocated resources being captured by the child process and stay alive forever.

This can result in a gigabyte-scale memory leak when you load checkpoint on CPU memory or do preprocessing of large data or initialize large model on CPU before calling `Task.init()`.

I faced this issue when using https://github.com/mosaicml/composer library. Callbacks such as logging are initialized after checkpoint is loaded/model initialized into CPU memory. After transferring the model to the GPU, CPU memory is still in use.

## To reproduce
```python
import torch
import time
import gc
from clearml import Task

leaked_tensor = torch.ones(4, 1024, 1024, 1024)

task = Task.init(project_name="leaked_tensor_project", task_name="debug mem leak")

del leaked_tensor

while True:
    collected = gc.collect()
    print(f"{collected=}")
    time.sleep(1)
```

## Environment
clearml==2.1.3

Edition-X:
## Summary

Fixes #1556 - Memory leak when resources are allocated before `Task.init()`

This PR switches from `fork` to `spawn` for background subprocess creation to prevent child processes from capturing pre-allocated memory from the parent process.

## Problem

When `os.fork()` was used to create the background monitoring subprocess, any large objects (models, datasets, tensors) allocated **before** `Task.init()` were captured in the child process's copy-on-write memory space and never released, causing memory leaks of hundreds of MB to multiple GB.

### Root Cause
The leak occurred in `clearml/utilities/process/mp.py:665`:
```python
BackgroundMonitor._main_process = os.fork()
```

## Solution

Switch to `spawn` context for multiprocessing, which starts a fresh Python interpreter instead of forking, preventing memory capture.

## Changes

- Use spawn context for subprocess creation on all platforms
- Update `SafeEvent` to use spawn context  
- Update process locks to use spawn context
- Enable subprocess on macOS (was Linux-only)
- Remove unused multiprocessing imports
- Add comment explaining atexit callback trade-off

### Linux (Docker - python:3.10-slim)
**Before Fix:**
- Child process: 552.10 MB ✗

**After Fix:**
- Child process: 9.41 MB ✓
- **Memory saved: ~542 MB**

### macOS (ARM/Apple Silicon)  
**Before Fix:**
- Child process: ~500 MB ✗

**After Fix:**
- Child process: 14.58 MB ✓
- **Memory saved: ~485 MB**

## Impact

This fix prevents gigabyte-scale memory leaks in common ML workflows:
- Loading large models (ResNet, BERT, GPT, LLMs)
- Pre-processing datasets into memory
- Frameworks initializing before ClearML (MosaicML Composer, PyTorch Lightning)
- Transfer learning workflows
- Any scenario with large pre-allocated objects

## Backward Compatibility

Using `spawn` instead of `fork` should be transparent to users:
- The child process only runs ClearML's internal background tasks (resource monitoring, uploads)
- No user code executes in the child process
- Aligns with Python 3.14+ direction (default spawn on macOS)
- Industry best practice (Gunicorn, PyTorch, TensorFlow use spawn)

The existing `__start_subprocess_forkprocess` method already handles the atexit callback concerns mentioned in the original TODO comment via daemon subprocess workarounds.

## Code Quality

✓ Passes flake8 linter per repository guidelines
✓ Tested on Linux and macOS
✓ No breaking changes to user-facing APIs

### DIFF
--- a/clearml/backend_interface/task/task.py
+++ b/clearml/backend_interface/task/task.py
@@ -95,7 +95,9 @@ class Task(IdObjectBase, AccessMixin, SetupUploadMixin):
 
     _store_diff = deferred_config("development.store_uncommitted_code_diff", False)
     _store_remote_diff = deferred_config("development.store_code_diff_from_remote", False)
-    _report_subprocess_enabled = deferred_config("development.report_use_subprocess", sys.platform == "linux")
+    # Enable subprocess on both Linux and macOS (darwin)
+    # We use spawn instead of fork to avoid memory leak (issue #1556)
+    _report_subprocess_enabled = deferred_config("development.report_use_subprocess", sys.platform in ("linux", "darwin"))
     _force_use_pip_freeze = deferred_config(
         multi=[
             ("development.detect_with_pip_freeze", False),
--- a/clearml/utilities/process/mp.py
+++ b/clearml/utilities/process/mp.py
@@ -3,7 +3,7 @@ import pickle
 import struct
 import sys
 from functools import partial
-from multiprocessing import Process, Semaphore, Event as ProcessEvent
+from multiprocessing import Semaphore
 from threading import Thread, Event as TrEvent, RLock as ThreadRLock
 from time import sleep, time
 from types import TracebackType
@@ -40,7 +40,8 @@ except ImportError:
 
 
 class _ForkSafeThreadSyncObject:
-    __process_lock = get_context("fork" if sys.platform == "linux" else "spawn").Lock()
+    # Use spawn context to avoid fork-based memory leak (issue #1556)
+    __process_lock = get_context("spawn").Lock()
 
     @classmethod
     def _inner_lock(cls) -> None:
@@ -454,7 +455,9 @@ class SafeEvent:
     __thread_pool = SingletonThreadPool()
 
     def __init__(self) -> None:
-        self._event = ProcessEvent()
+        # Use spawn context to avoid fork-based memory leak (issue #1556)
+        ctx = get_context("spawn")
+        self._event = ctx.Event()
 
     def is_set(self) -> bool:
         return self._event.is_set()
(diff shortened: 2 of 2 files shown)


Edition-X:
## Also Fixes Issue #1522

This PR also resolves issue #1522 (os.fork() DeprecationWarning in Python 3.12+).

By switching from `os.fork()` to `spawn`-based multiprocessing, we eliminate:
- ✅ The DeprecationWarning raised in Python 3.12+
- ✅ Thread safety issues from forking in multi-threaded contexts
- ✅ Risk of deadlocks from copied locks

One fix, two issues resolved!

Fixes #1556
Fixes #1522

patrickdasilva-clearml:
Dear @Edition-X ,

thank you for opening this PR. However, I must decline it, because as was discussed in the issue thread, the issue is that `Task.init` must be called as early as possible, as this is by design. Users should not allocate anything except defining functions/classes and making import statements before `Task.init`, precisely to avoid this memory leak (among other things).

We updated the docs to make it more clear, although if you think the docs are not clear enough, feel free to suggest something. For the moment, I will close this PR.

Edition-X:
> Dear @Edition-X ,
> 
> thank you for opening this PR. However, I must decline it, because as was discussed in the issue thread, the issue is that `Task.init` must be called as early as possible, as this is by design. Users should not allocate anything except defining functions/classes and making import statements before `Task.init`, precisely to avoid this memory leak (among other things).
> 
> We updated the docs to make it more clear, although if you think the docs are not clear enough, feel free to suggest something. For the moment, I will close this PR.

Thanks for getting back to me. Yea I think I was a bit to hasty with a fix here, the discussion makes sense.
If there is something I can contribute to help out I'm more than happy to and I would love to get involved.

patrickdasilva-clearml:
@Edition-X of course! If you're willing to contribute, browsing the issues section is one way. Another way is to join our Slack community (you can find the link [here](https://clear.ml/community)) which has plenty of active users. Thanks again!

PR #1559, opened 2026-03-16
Fix timezone handling in datetime_from_isoformat

jday1:
## Summary
- make `datetime_from_isoformat` preserve timezone offsets when parsing ISO strings
- support `Z` suffix as UTC
- default naive parsed datetimes to UTC tzinfo
- widen type annotation to `Optional[Union[str, datetime]]`

## Why
Scheduler paths expect timezone-aware datetimes (UTC). The previous parser dropped timezone offset information and could return naive datetimes after (de)serialization, causing inconsistent behavior and potential datetime comparison errors.

## Validation
- manual checks in project env:
  - `datetime_from_isoformat('1970-01-01T00:00:00')` -> UTC-aware
  - `datetime_from_isoformat('1970-01-01T00:00:00+00:00')` -> UTC-aware
  - `datetime_from_isoformat('1970-01-01T00:00:00Z')` -> UTC-aware

### DIFF
--- a/clearml/automation/scheduler.py
+++ b/clearml/automation/scheduler.py
@@ -158,6 +158,7 @@ class ScheduleJob(BaseScheduleJob):
                 day=int(self.day or 1),
                 hour=int(self.hour or 0),
                 minute=int(self.minute or 0),
+                tzinfo=timezone.utc,
             )
             if self.weekdays:
                 self._next_run += relativedelta(weekday=self.get_weekday_ord(self.weekdays[0]))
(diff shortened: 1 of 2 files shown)


patrickdasilva-clearml:
Hi @jday1 , thanks for contributing! We have delivered a fix for the underlying issue #1514 in `clearml==2.1.5`, so we don't need this PR anymore, which is why I'm closing it. Let us know if you still wish to contribute by opening another PR.

PR #1562, opened 2026-03-17
fix(hydra): reload task data before fetching remote parameters

### LINKED ISSUE #1536: Hydra overrides are not applied if using uv as the package manager
## Describe the bug
In the following workflow:
- Run a task locally
- Clone in the UI
- Change a configuration under Configuration > OmegaConf
- Set `_allow_omegaconf_edit_ = True` under Configuration > Hydra
- Enqueue

I seem to be getting different results between pip and uv.

With pip, the parameters are successfully overridden. With uv, they fall back to whatever was in the original config.yaml. 

## To reproduce
Clone this repo: https://github.com/Anjum48/uv-hydra-testing

In the main branch, create a pip venv, and run `entrypoint.py`:

```
python -m venv .venv_pip
.\.venv_pip\Scripts\Activate.ps1
pip install -r .\requirements.txt
python .\entrypoint.py
```

Then override the `run_name` parameter (e.g. `this is from pip`) as described above. You will see this in the console after enqueuing:

```
Hello, Hydra and ClearML!
{'run': {'timestamp': '${now:%Y%m%d-%H%M%S}', 'run_name': 'this is from pip'}}
```

Now in the `uv-testing` branch, create a uv venv and repeat:
```
uv venv
.\.venv\Scripts\Activate.ps1
uv pip install -r .\requirements.txt
python .\entrypoint.py
```

Then override the `run_name` parameter (e.g. `this is from uv`) as described above. You will see this in the console:

```
Hello, Hydra and ClearML!
{'run': {'timestamp': '${now:%Y%m%d-%H%M%S}', 'run_name': 'Default_Run_Name'}}
```

## Expected behaviour
What is the expected behaviour? What should've happened but didn't?

In the final step we should have seen:

```
Hello, Hydra and ClearML!
{'run': {'timestamp': '${now:%Y%m%d-%H%M%S}', 'run_name': 'this is from uv'}}
```

What actually happened is that it defaulted to `config\config.yaml`

## Environment
* Server type: self hosted
* ClearML SDK Version: 2.1.3
* ClearML Server Version: WebApp: 2.3.0-705 • Server: 2.3.0-705 • API: 2.34
* OS: Windows local, Linux server
## Related Discussion

BillionClaw:
## Related Issue

Fixes #1536

## Patch Description

When running remotely with Hydra, the task data needs to be reloaded to ensure we have the latest parameters from the backend. This is important because the task may have been modified (e.g., cloned and edited in the UI) before being enqueued for remote execution.

The issue manifested as Hydra overrides not being applied when using uv as the package manager. This was because the task was reading stale cached data that did not include the user's configuration changes made in the ClearML UI (such as setting `_allow_omegaconf_edit_ = True` and modifying the OmegaConf section).

## Root Cause

In `_patched_hydra_run`, when running remotely, the code retrieves the task and immediately tries to get parameters like `_allow_omegaconf_edit_`. However, the task object may have stale cached data from when it was first initialized. Without calling `reload()` to fetch the latest data from the backend, the code would use outdated parameter values, causing the Hydra configuration overrides to be ignored.

## Fix

Added a `reload()` call after getting the task object in `_patched_hydra_run` to ensure we have the latest task data from the backend before reading parameters.

### DIFF
--- a/clearml/binding/hydra_bind.py
+++ b/clearml/binding/hydra_bind.py
@@ -107,6 +107,10 @@ class PatchHydra:
                 from ..task import Task
 
                 PatchHydra._current_task = Task.get_task(task_id=get_remote_task_id())
+            # Reload task data to ensure we have the latest parameters from the backend
+            # This is important when running remotely as the task may have been modified
+            # (e.g., cloned and edited) before being enqueued
+            PatchHydra._current_task.reload()
             # get the _parameter_allow_full_edit casted back to boolean
             connected_config = {}
             connected_config[PatchHydra._parameter_allow_full_edit] = False


patrickdasilva-clearml (commented):
Hi @BillionClaw , thank you for submitting this PR. I have two questions:
- Does your PR explain why this problem happens with `uv`, but not when using other approaches (e.g. with `pip`)? How do you explain that?
- How have you tested this change? I have no idea how to confirm based on the PR alone.

P.S. : (I'm assuming from your username that you used AI to generate this fix; if my assumption is incorrect, please dismiss this comment.) It's great to use AI for coding topics (I'm doing it most of the time as well), but you're still interacting with people here. We need more than just the code and the claim that things are fixed to feel confident about merging changes.

lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

patrickdasilva-clearml:
@lgx0319 I asked the author a question, and they haven't replied. I've known them to use AI to open PRs so I prefer to wait for their reply before deciding to merge it or not. I'm keeping it open either until the underlying problem is solved or the author replies.

patrickdasilva-clearml:
This PR will be closed due to inactivity, and also doubting that the suggested changes actually solve the problem. Feel free to re-open the PR if you think it's still relevant.

PR #1585, opened 2026-03-25
Fix deprecation codecs.open() is deprecated. Use open() instead.

mscheltienne:
Fix deprecation in 3.14:

```
codecs.open() is deprecated. Use open() instead.
```

`open` with `encoding` has been available since Python 3.

### DIFF
--- a/clearml/utilities/pigar/__main__.py
+++ b/clearml/utilities/pigar/__main__.py
@@ -2,7 +2,6 @@
 
 from __future__ import print_function, division, absolute_import
 
-import codecs
 import os
 from typing import Callable, Set, Optional, Tuple, List, Any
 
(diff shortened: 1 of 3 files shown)


patrickdasilva-clearml:
Hi @mscheltienne ! Thanks for opening this PR. I've been concerned with deprecating old Python2-related features myself, and I hadn't spotted this one, so I appreciate it.

I'm gonna run some tests of my own on it to make sure this won't be problematic and get back to you on this PR tomorrow.

patrickdasilva-clearml:
Hi @mscheltienne , sorry for taking a few days to get back to you on this.

We decided not to merge the PR, but not because it's not correct (it is).
- In `setup.py`, we have short-term plans to migrate to a `pyproject.toml` format, where this code would go away anyway. So it's simpler to leave it as it is for now.
- In the other two files (`clearml/utilities/pigar/__main__.py` and `clearml/utilities/pyhocon/config_parser.py`) , we have an internal decision of not touching those specific files at all for reasons relevant to our business.

Feel free to reach out to us on our community Slack if you want to discuss this in more detail, and don't let that refrain you from taking initiative in the future, we very much support that.

PR #1590, opened 2026-04-05
Feat/external links upload

tabibon:
### `62997c3d` — SHA-256 hash comparison for external link change detection

- `LinkEntry` gains a `hash` field and persists it through `as_dict()` serialization
- `add_external_files` gets a `read_hash` flag; when set, SHA-256 is read from each object's custom cloud metadata and stored on the `LinkEntry`
- Change detection now prefers hash comparison over size when both sides carry a hash, falling back to size when hash is unavailable
- `StorageHelper.get_object_metadata` reads SHA-256 from custom blob metadata (`"sha256"` key); if missing, attempts `reload()` on the object
- `StorageHelper.list` and `StorageManager.list` / `get_metadata` thread `read_hash` through to the driver layer
- GCS and Azure upload paths store the SHA-256 in object custom metadata when `upload_hash` is present in `extra`; S3 stores it under `Metadata["sha256"]`
- S3 `list_container_objects` now gates the per-object HEAD call behind `read_hash` to avoid N+1 requests when hash is not needed

---

### `bed10556` — Subset file pulling in `get_local_copy`

- `get_local_copy` gains a `files` parameter accepting a list of relative paths; when set, only those files are downloaded
- Cache folder is suffixed with an 8-char SHA-256 hash of the sorted file list, keeping subset caches independent from each other and from the full-dataset cache
- New `_build_subset_chunk_selection` walks the full dependency graph and maps requested files to the chunk indices that contain them
- `files_of_interest` is threaded through `_merge_datasets`, `_extract_parent_datasets`, and `_verify_dataset_folder` to filter file entries, link entries, and copy operations
- For link-only datasets where no chunk-backed files match, falls back to full dependency resolution so link entries are still downloaded
- Default behaviour (`files=None`) is unchanged

---

### `9174775e` — Prune stale link entries in `sync_folder`; `upload_as_external_links` flag

**`sync_folder`**
- The same local-existence filter already applied to `_dataset_file_entries` is now also applied to `_dataset_link_entries`; stale external-link records are removed when their local counterpart is absent
- Combined removed count (files + links) is passed to `update_changed_files` in a single call

**`upload(upload_as_external_links=True)`**
- New flag uploads each local file entry directly to `<output_url>/external_links/<dataset_id>/<relative_path>` via `StorageHelper` using a thread pool (respects `max_workers` and cloud-vs-local heuristic)
- After upload, `_dataset_file_entries` is cleared and `add_external_files` registers all uploaded objects as `LinkEntry` records
- The file's pre-computed SHA-256 is forwarded to `helper.upload` via `extra={"hash": ...}` for backends that store it as object metadata
- The existing zip/chunk path then runs on the now-empty file entries (no-op) and serializes normally


clearml/datasets/dataset.py
```diff
@@ -738,6 +755,48 @@ def upload(
...
+            "Uploading dataset files as external links: {}".format(
+                dict(
+                    show_progress=show_progress,
+                    verbose=verbose,
+                    output_url=output_url,
+                    compression=compression,
+                )
+            ),
+            print_console=False,
+        )
```
eugen-ajechiloae-clearml: this section has odd identation

clearml/datasets/dataset.py
```diff
@@ -3725,11 +3887,11 @@ def _add_external_files(
                         relative_path=relative_path,
                         parent_dataset_id=self._id,
                         size=size,
+                        hash=hash_val,
                     )
                     num_modified += 1
-                elif (
-                    relative_path in self._dataset_link_entries
-                    and self._dataset_link_entries[relative_path].size != size
+                elif hash_val and existing_link.hash and existing_link.hash != hash_val or (
```
eugen-ajechiloae-clearml: this condition is hard to read - let's use some parentheses around this section
```
hash_val and existing_link.hash and existing_link.hash != hash_val
```

clearml/datasets/dataset.py
```diff
@@ -2748,9 +2827,14 @@ def _merge_datasets(
             part=0 -> chunks[0,5], part=1 -> chunks[1,6], part=2 -> chunks[2,7], part=3 -> chunks[3, ]
         :param max_workers: Number of threads to be spawned when merging datasets. Defaults to the number
             of logical cores.
+        :param files_of_interest: Optional set of relative file paths. When provided, only those files
+            (and the chunks that contain them) are downloaded. The cache folder is suffixed with a short
+            hash derived from this set.
 
         :return: the target folder
         """
+        import hashlib
```
eugen-ajechiloae-clearml: we can import this at the top - hashlib is a standard library

eugen-ajechiloae-clearml (changes requested):
Hello! Added a few comments here. Any chance you also have (some) test script(s) for the changes?

tabibon:
@eugen-ajechiloae-clearml i resolved your comments, 
added unit tests and e2e test, i used out clearml deployment, u might need to change clearml_project and output_bucket

eugen-ajechiloae-clearml:
Hi @tabibon ! Thank you for the changes. One more thing regarding the tests: can you please create in repo root a directory called `tests` and a subdirectory called `test_datasets` (so `tests/test_datasets`). This will keep the clearml package clean and help us migrate the tests to our internal testing process

patrickdasilva-clearml:
We have discussed this PR with the author personally and integrated their tests in our internal testing processes. We needed to make some changes so we opened [another PR on the same topic](https://github.com/clearml/clearml/pull/1611). That PR was merged, so we don't need this PR anymore. Therefore, we are closing it, but its contents were accepted.

Thanks @tabibon for the contribution and collaboration!

PR #1599, opened 2026-04-16
Add pickle file integrity verification for pd.DataFrame pickled artifacts

patrick-da-silva-ST:
## Related Issue \ discussion
There are some CVE reports claiming that we should add an integrity check to `pandas.DataFrame` pickled artifacts on retrieval.

## Patch Description
This adds a hash verification to make sure that `pandas.DataFrame` pickled artifacts haven't been tampered with. The hash is stored on the artifact when creating it, and this PR re-computes it from the file and checks it against the stored hash. This was done before for pickled artifacts, but `pandas.DataFrame` pickled artifacts are pickled with `df.to_pickle()` and get read with `pd.read_pickle`, so the process is slightly different.

### DIFF
--- a/clearml/binding/artifacts.py
+++ b/clearml/binding/artifacts.py
@@ -48,6 +48,11 @@ if TYPE_CHECKING:
     import pandas
 
 
+class ArtifactIntegrityError(ValueError):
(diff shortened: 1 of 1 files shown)


patrickdasilva-clearml:
LGTM

PR #1604, opened 2026-04-24
fix: adjust docstrings in clearml/task.py

patrick-da-silva-ST:
## Patch Description
We're making internal adjustments to the docstrings in `clearml/task.py` for the documentation at `clear.ml/docs`.


patrickdasilva-clearml (approved):
LGTM

PR #1605, opened 2026-04-25
Refactor of clearml/storage/util.py

patrick-da-silva-ST:
## Patch Description
This is a refactor of `clearml/storage/util.py` to improve its readability. There was also an edge case which was found during the refactor, namely that `format_size` would render `999.995 KB` as `1000 KB` (undesired behavior) instead of `1 MB` (desired behavior). This was fixed and made obvious in the refactor.

## Testing Instructions
Tests have been ran internally.


patrickdasilva-clearml:
LGTM

PR #1606, opened 2026-04-26
Refactor/clearml storage filepaths

patrick-da-silva-ST:
## Patch Description
This refactors filepath-related functionality in its own file.
Also, a backwards-compatibility bug was introduced [in the last PR](https://github.com/clearml/clearml/pull/1605) related to the location of external imports, and it was fixed (see second commit). This bug has not been introduced in a release yet, so no issues there.


patrickdasilva-clearml:
LGTM

PR #1610, opened 2026-04-30
Refactor/clearml storage archive update

patrick-da-silva-ST:
## Patch Description
This PR adds improvements to the `clearml/storage` directory.
- Refactor of `clearml/storage/util.py` to produce `clearml/storage/archive.py` for functions related to extracting `.tar.gz`, `.tgz` and `.zip` files contents.
- Removal of redundant code from the `create_zip_directories` function. This code was duplicated from the `zipfile` package from the Python standard library with useless `try-except` safeguards, because the call to `ZipFile().extractall` done right after would execute the same code without the safeguard, resulting in the same errors.
- Minor improvements to `examples/hyperdatasets`.

### DIFF
--- a/clearml/storage/archive.py
+++ b/clearml/storage/archive.py
@@ -1,0 +1,110 @@
+import os
+import tarfile
+from typing import Optional, Union, Any
+from pathlib import Path
+
+from .filepaths import is_within_directory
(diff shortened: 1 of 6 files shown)


patrickdasilva-clearml:
LGTM

PR #1611, opened 2026-04-30
Hash comparison, subset file pulling, prune stale link entries

patrick-da-silva-ST:
## Related Issue \ discussion
See [this PR](https://github.com/clearml/clearml/pull/1590) for the original implementation attempt.

#### SHA-256 hash comparison for external link change detection
- `LinkEntry` gains a `hash` field and persists it through `as_dict()` serialization
- `add_external_files` gets a `read_hash` flag; when set, SHA-256 is read from each object's custom cloud metadata and stored on the `LinkEntry`
- Change detection now prefers hash comparison over size when both sides carry a hash, falling back to size when hash is unavailable
- `StorageHelper.get_object_metadata` reads SHA-256 from custom blob metadata (`"sha256"` key); if missing, attempts `reload()` on the object
- `StorageHelper.list` and `StorageManager.list` / `get_metadata` thread `read_hash` through to the driver layer
- GCS and Azure upload paths store the SHA-256 in object custom metadata when `upload_hash` is present in extra; S3 stores it under `Metadata["sha256"]`
- S3 `list_container_objects` now gates the per-object HEAD call behind `read_hash` to avoid N+1 requests when hash is not needed

#### Subset file pulling in get_local_copy
- `get_local_copy` gains a `files` parameter accepting a list of relative paths; when set, only those files are downloaded
- Cache folder is suffixed with an 8-char SHA-256 hash of the sorted file list, keeping subset caches independent from each other and from the full-dataset cache
- New `_build_subset_chunk_selection` walks the full dependency graph and maps requested files to the chunk indices that contain them
- `files_of_interest` is threaded through `_merge_datasets`, `_extract_parent_datasets`, and `_verify_dataset_folder` to filter file entries, link entries, and copy operations
- For link-only datasets where no chunk-backed files match, falls back to full dependency resolution so link entries are still downloaded
- Default behaviour (`files=None`) is unchanged

#### Prune stale link entries in `sync_folder`; `upload_as_external_links` flag
`sync_folder`
- The same local-existence filter already applied to `_dataset_file_entries` is now also applied to `_dataset_link_entries`; stale external-link records are removed when their local counterpart is absent
- Combined removed count (files + links) is passed to `update_changed_files` in a single call
`upload(upload_as_external_links=True)`
- New flag uploads each local file entry directly to `<output_url>/external_links/<dataset_id>/<relative_path>` via `StorageHelper` using a thread pool (respects `max_workers` and cloud-vs-local heuristic)
- After upload, `_dataset_file_entries` is cleared and `add_external_files` registers all uploaded objects as `LinkEntry` records
- The file's pre-computed SHA-256 is forwarded to `helper.upload` via `extra={"hash": ...}` for backends that store it as object metadata
- The existing zip/chunk path then runs on the now-empty file entries (no-op) and serializes normally

## Testing
We integrated the tests from [the original PR](https://github.com/clearml/clearml/pull/1590) in our internal testing processes.

### DIFF
--- a/clearml/datasets/dataset.py
+++ b/clearml/datasets/dataset.py
@@ -10,7 +10,7 @@ from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy, copy
 from multiprocessing.pool import ThreadPool
 from tempfile import mkdtemp
-from typing import Union, Optional, Sequence, List, Dict, Mapping, Tuple, TYPE_CHECKING, Any
+from typing import Union, Optional, Sequence, List, Dict, Mapping, Tuple, TYPE_CHECKING, Any, Set
 from zipfile import ZIP_DEFLATED
 from collections import deque
 
@@ -115,6 +115,7 @@ class LinkEntry:
             relative_path=self.relative_path,
             parent_dataset_id=self.parent_dataset_id,
             size=self.size,
+            hash=self.hash,
         )
 
 
@@ -482,6 +483,7 @@ class Dataset:
         recursive: bool = True,
         verbose: bool = False,
         max_workers: Optional[int] = None,
+        read_hash: bool = False,
     ) -> int:
         """
         Adds external files or folders to the current dataset.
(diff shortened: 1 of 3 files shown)


patrickdasilva-clearml:
LGTM

Issue #1547, opened 2026-03-04
Improve readability by replacing `.format` string methods by `f`-strings


patrickdasilva-clearml:
## Proposal Summary
We support Python 3.6+, and `f`-strings were introduced in Python 3.6. We should stop using the following patterns and replace them by `f`-strings for readability:
- `"%s is %d" % (str_value, int_value)`
    This is very hard to read.
- `"{0} is {1}".format(name, age)`
    This is hard to read, but still necessary in the case the string to format (`"{0} is {1}"` in the example) is provided dynamically; in most cases, `.format` usage is not necessary and `f`-strings do the job better.

## Motivation
We support Python 3.6+, so we have no reason to keep our code with older patterns if newer patterns do a better job.

patrickdasilva-clearml:
We merged one PR on the topic but there are many more to go. Sorry for the accidental closure-reopening of this issue.

PR #1615, opened 2026-05-05
refactor: move .format calls to f-strings

patrick-ds-spiritech:
## Related Issue \ discussion
See issue #1547 

## Patch Description
This targets the issue in the `clearml/storage` folder. Does not affect runtime behavior.


patrickdasilva-clearml:
LGTM

PR #1617, opened 2026-05-08
feat: Move shutdown process under try-finally

patrick-ds-spiritech:
## Patch Description
We received a report that the `clearml/utilities/gpu/gpustat.py` has a problem when the GPU fails during the loop in `GPUStatCollection._new_query_nvidia`. This PR addresses that issue to ensure that the conditional shutdown process never fails to occur if something happens at the hardware level.


patrickdasilva-clearml:
LGTM

PR #1618, opened 2026-05-08
Feat/block pickled artifacts retrieval

patrick-ds-spiritech:
## Patch Description
This gives users the ability to protect themselves against `pickle`-related RCEs by blocking the download and execution of pickled artifacts via configuration. It is an "opt-out" feature for unpickling, so the default behavior does not change.


patrickdasilva-clearml:
LGTM

PR #1619, opened 2026-05-11
Add numpy pickled artifacts hash integrity check on retrieval

patrick-ds-spiritech:
## Patch Description
We recently added a hash integrity check when unpickling pandas dataframe artifacts. This PR does the same for numpy array pickled artifacts.


patrickdasilva-clearml:
LGTM

PR #1620, opened 2026-05-11
Add vulnerability check: import_offline_session zipfile.extractall call

patrick-ds-spiritech:
## Patch Description
We previously extracted vulnerability check during `.extractall` method calls on zipfiles but one such instance of that call was left behind. This PR resolves that issue.


patrickdasilva-clearml:
LGTM

PR #1621, opened 2026-05-12
Rename `block_pickled_artifacts` to `block_unsafe_artifacts`

patrick-ds-spiritech:
## Patch Description
A new argument to the `Artifact.get` method was recently added to prevent the download of pickled artifacts if they are deemed unsafe due to known RCE vulnerabilities related to unpickling. To make it easier to preserve backwards compatibility in the future, we are renaming this argument.


patrickdasilva-clearml:
LGTM

PR #1623, opened 2026-05-14
fix: add path traversal check in Dataset._merge_datasets

patrickdasilva-clearml:
## Patch Description
This PR adds validation to the `Dataset._merge_datasets` method against path-traversal attacks.


patrickdasilva-clearml:
LGTM

PR #1625, opened 2026-05-15
feat: Add dataset.status in Dataset.list_datasets return values

patrickdasilva-clearml:
## Patch Description
This PR adds `dataset["status"]` in the return value of the `Dataset.list_datasets` method. This can be used to then further filter that list of datasets and make more detailed queries via `Dataset.get` afterwards.


patrickdasilva-clearml:
LGTM

PR #1693, opened 2026-08-16
Fix nondeterministic type validation messages

### LINKED ISSUE #1692: `DataModel.assert_isinstance` can produce nondeterministic type-error messages
## Describe the bug

Hi! I found a small reproducibility issue in the SDK's array-type validation. When `DataModel.assert_isinstance(..., is_array=True)` rejects a collection, it formats the observed type names using:

```python
', '.join(set(type(x).__name__ for x in value))
```

Because the set contains strings, its iteration order can depend on Python's per-process hash seed. A mixed invalid collection can therefore produce the same `TypeError` with its type names in different orders, across otherwise equivalent interpreter runs.

This doesn't affect the validation result or any successful request. However, it does make the user-facing error text, logs, snapshots, and assertions against that text, kind of unstable.

## To reproduce

Save this as `reproduce.py` in an environment containing ClearML:

```python
from clearml.backend_api.session.datamodel import DataModel

try:
    DataModel.assert_isinstance(
        [1, "a", None],
        "values",
        float,
        is_array=True,
    )
except TypeError as error:
    print(error)
```

Then run it in fresh interpreters with several hash seeds:

```console
PYTHONHASHSEED=1 python reproduce.py
PYTHONHASHSEED=2 python reproduce.py
PYTHONHASHSEED=3 python reproduce.py
PYTHONHASHSEED=4 python reproduce.py
```

The final `got ...` list can appear in different orders. In a local check over 16 hash seeds, the current expression produced all six permutations of `NoneType`, `int`, and `str`.

## Expected behaviour

The same invalid value should produce a stable validation message across interpreter processes. Sorting the deduplicated type-name strings before joining them would establish a deterministic order without changing which types are reported.

## Environment
* Server type: Not applicable; this occurs in SDK-side validation.
* ClearML SDK Version: `a16678313800a431ef6757be58a3858624077602`
* ClearML Server Version: Not applicable.
* Python version: 3.14.3 for the reproduction; the behavior follows ordinary Python string-hash randomization rather than a 3.14-specific change.
* OS: MacOS
## Related Discussion
None.

marcus-campbell:
## Related Issue \ discussion
Fixes #1692<br/>

## Patch Description
This sorts the deduplicated type names in `DataModel.assert_isinstance(..., is_array=True)` before including them in the raised `TypeError`.

The current set iteration order can depend on Python's per-process hash seed, so the same mixed invalid collection may produce different validation text across interpreter runs. Sorting keeps the message deterministic without changing which types are reported, or affecting successful validation.<br/>

## Testing Instructions
* Run the helper with the same mixed invalid value under 16 `PYTHONHASHSEED` values. The previous expression produced all six possible orderings of three type names; this change produces one stable message.
* The modified file passes ClearML's documented Flake8 configuration:

  ```console
  flake8 --max-line-length=120 --statistics --show-source --extend-ignore=E501 clearml/backend_api/session/datamodel.py
  ```
* The documented repository-wide Flake8 command reports the same 46 pre-existing `E203` findings on this branch and on untouched `master`; none are in the modified file.

## Other Information
This changes only the ordering of type names in an already-failing validation path.

### DIFF
--- a/clearml/backend_api/session/datamodel.py
+++ b/clearml/backend_api/session/datamodel.py
@@ -123,7 +123,7 @@ class DataModel:
         if not all(isinstance(x, expected) for x in value):
             raise TypeError(
                 f"Expected {field_name} of type list[{expected}], "
-                f"got {', '.join(set(type(x).__name__ for x in value))}"
+                f"got {', '.join(sorted(set(type(x).__name__ for x in value)))}"
             )
 
     @staticmethod


patrickdasilva-clearml (approved):
Thanks for the changes! We should follow this practice in the future when producing error messages for consistency.

